PR #1, opened 2020-04-04
Fix the build given the repo separation.

peterhuene:
This commit ports the files from the Wasmtime repo.


peterhuene:
I'll push up a commit to port the CI and then some additional documentation.

radu-matei:
Would there be any value in keeping the original history from the Wasmtime repo?

Otherwise, this looks great!

peterhuene:
There probably is value, I'll see about transferring the history as well.

PR #12, opened 2020-04-22
Fix CI.

peterhuene:
This PR fixes CI as currently Wasmtime will panic if non-funcref tables are
used (not currently implemented).

### DIFF
--- a/tests/Modules/TableExports.wat
+++ b/tests/Modules/TableExports.wat
@@ -1,6 +1,6 @@
 (module
   (table $t0 1 10 funcref)
-  (table $t1 10 anyref)
+  (table $t1 10 funcref)
   (table $t2 100 1000 funcref)
   (export "table1" (table $t0))
   (export "table2" (table $t1))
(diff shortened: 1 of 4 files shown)


peterhuene:
Before we merge this, I think this is a regression in Wasmtime; I'll confirm.

peterhuene:
This appears to have been was regressed by bytecodealliance/wasmtime#1524.

Previously we were just cloning an internal table type in `Table::ty` and now we're [calling `TableType::from_wasmtime_table`](https://github.com/bytecodealliance/wasmtime/pull/1524/files#diff-db60ba5490a564d3634db61c9ac0d399R344) which panics if the element type is not funcref.

@sunfishcode previously the C API was able to describe anyref table imports and exports, but now we hit the assertion in `TableType::from_wasmtime_table`.  Given that anyref tables aren't really implemented yet in Wasmtime, how should we fix this?  Should the .NET tests simply just stick to funcref tables in the test modules until Wasmtime reference types implementation is more complete or should we relax the assertion such that anyref is valid?

Wasmtime will `unimplemented!` panic with the following module despite reference types being enabled:

```wat
(module
  (table $t 1 anyref)
)
```

peterhuene:
I'm going to merge this for now and we can revisit when more of the reference types proposal is implemented.

PR #15, opened 2020-06-06
Improvements for error messages and traps.

### LINKED ISSUE #7: Allow reporting errors when loading wasm modules into .NET
#### Feature

When loading wasm modules into .NET with wasmtime, an exception is raised if a module is invalid. This exception doesn't provide any details as to what went wrong, however, making it hard to fix the module.

#### Implementation

The .NET implementation relies on the `wasm_module_new` function.
https://github.com/bytecodealliance/wasmtime/blob/121bbd36568c506a39b2dd97a95dd2b963960eb7/crates/c-api/src/lib.rs#L828

This function currently discards the error information with a match expression:

https://github.com/bytecodealliance/wasmtime/blob/121bbd36568c506a39b2dd97a95dd2b963960eb7/crates/c-api/src/lib.rs#L836

Allowing this information to be captured in a callback or in a mutable pointer to a string that could capture the Debug derive for the `anyhow::Error` would be helpful in diagnosing errors that occur when loading modules.

### LINKED ISSUE #11: How to see stack traces?
Maybe Wasmtime dotnet wrapper includes old wastime without stack traces support?

Stack traces can help for debugging, for example "unreachable" functions.

peterhuene:
This PR improves error messages for linking errors, module load errors, and provides frame details for `TrapException`.

Fixes #7.
Fixes #11.


peterhuene:
Test failures are due to an upstream Wasmtime bug.

PR #19, opened 2020-06-16
Added table exports to Externs

### LINKED ISSUE #17: No support for tables in Instance.Externs
If Module.Exports.Tables contains any values, the exception NotSupportedException with the message "Unsupported extern type." is thrown in the Externs constructor when Host.Instantiate is called

jlgaffney:
Fixes #17

Include table exports into the Externs instance

### DIFF
--- a/src/Externs/ExternTable.cs
+++ b/src/Externs/ExternTable.cs
@@ -1,0 +1,40 @@
+using System;
+using Wasmtime.Exports;
+
+namespace Wasmtime.Externs
+{
+    /// <summary>
+    /// Represents an external (instantiated) WebAssembly table.
+    /// </summary>
+    public class ExternTable
+    {
+        internal ExternTable(TableExport export, IntPtr table)
+        {
+            _export = export;
+            _table = table;
+        }
+
+        /// <summary>
+        /// The name of the WebAssembly memory.
+        /// </summary>
+        public string Name => _export.Name;
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


tests/TableExportsTests.cs
```diff
@@ -37,6 +37,33 @@ public void ItHasTheExpectedNumberOfExportedTables()
             GetTableExports().Count().Should().Be(Fixture.Module.Exports.Tables.Count);
         }
 
+        [Fact]
+        public void ItCreatesExternsForTheGlobals()
```
peterhuene, line 41: Should this be `ItCreatesExternsForTheTables`?

src/Externs/ExternTable.cs
```diff
@@ -0,0 +1,40 @@
...
+    public class ExternTable
+    {
+        internal ExternTable(TableExport export, IntPtr table)
+        {
+            _export = export;
+            _table = table;
+        }
+
+        /// <summary>
+        /// The name of the WebAssembly memory.
```
peterhuene, line 18:
```suggestion
        /// The name of the WebAssembly table.
```

jlgaffney:
Ah yes, sorry about that

tests/TableExportsTests.cs
jlgaffney: Fixed it, should be good now

peterhuene:
@jlgaffney thanks for this!

PR #21, opened 2020-06-25
Add the Store type back to the API.

### LINKED ISSUE #9: Details on "Reimplement the C# API" for 0.15

peterhuene:
This is a breaking API change.

The `Host` class was split back into a `Store` type, where loading of modules
will occur.

This allows for loading modules into a single store and defining one or more
hosts that instantiate the modules without having to load the modules again.

Fixes #9.

### DIFF
--- a/docs/articles/intro.md
+++ b/docs/articles/intro.md
@@ -65,15 +65,15 @@ namespace Tutorial
     {
         static void Main(string[] args)
         {
-            using var host = new Host();
+            using var store = new Store();
+            using var module = host.LoadModuleText("hello", "(module (func $hello (import \"\" \"hello\")) (func (export \"run\") (call $hello)))");
+            using var host = new Host(store);
 
             host.DefineFunction(
                 "",
                 "hello",
                 () => Console.WriteLine("Hello from C#!")
             );
-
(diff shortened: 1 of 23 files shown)


docs/articles/intro.md
```diff
@@ -65,16 +65,16 @@ namespace Tutorial
     {
         static void Main(string[] args)
         {
-            using var host = new Host();
+            using var store = new Store();
+            using var module = host.LoadModuleText("hello", "(module (func $hello (import \"\" \"hello\")) (func (export \"run\") (call $hello)))");
```
yurydelendik, line 69:
```suggestion
            using var module = store.LoadModuleText("hello", "(module (func $hello (import \"\" \"hello\")) (func (export \"run\") (call $hello)))");
```

peterhuene: Good eye! 👍

PR #22, opened 2020-06-26
Leave the supplied stream open in `LoadModuleText`

yvt:
This PR modifies `Host.LoadModuleText(string, Stream)` to leave the supplied stream open like `Host.LoadModule(string, Stream)` does.

It also includes tests to check this behavior for both of `LoadModule` and `LoadModuleText`.


peterhuene:
@yvt Thanks for fixing this!

PR #24, opened 2020-07-01
Use a different store for the host.

peterhuene:
The host was using the same store as was given in the constructor for the
underlying linker and defined memories/functions/globals.

As a result, a memory defined in a host, for example, would not get freed when
the host (and all instances) were disposed.

This prevents the use case where a module gets loaded once into a store and then
a host is created to instantiate the module.  If the operation to create the
host and instantiate the module is performed in a loop (e.g. a benchmark), then
memory continues to grow as the original store is never released.

### DIFF
--- a/src/Host.cs
+++ b/src/Host.cs
@@ -24,9 +24,10 @@ namespace Wasmtime
                 throw new ArgumentNullException(nameof(store));
             }
 
-            _store = store;
+            // Create a separate store for the host
+            _store = Interop.wasm_store_new(store.EngineHandle);
 
-            var linker = Interop.wasmtime_linker_new(store.Handle);
+            var linker = Interop.wasmtime_linker_new(_store);
             if (linker.IsInvalid)
             {
                 throw new WasmtimeException("Failed to create Wasmtime linker.");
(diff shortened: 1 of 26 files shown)


peterhuene:
Looks like the benchmark silently failed on Windows because it couldn't find Wasmtime.dll.  Fixing that now.

src/Host.cs
```diff
@@ -24,9 +24,10 @@ public Host(Store store)
                 throw new ArgumentNullException(nameof(store));
             }
 
-            _store = store;
+            // Create a separate store for the host
+            _store = Interop.wasm_store_new(store.EngineHandle);
```
yurydelendik, line 28: The only place I don't fully understand is why `Host`'s parameters have `Store`. Ideally there shall be an engine, having `Store` to create a new engine looks inefficient.

At this point, I wonder if `Store` needs to be removed/encapsulated by `Host` (and `StoreBuilder` replaced by `Engine)?

peterhuene: I could add back the `Engine` abstraction.  I agree that passing in a store just to get the internal engine here is confusing.

Let me do that.

peterhuene:
@yurydelendik I've pushed up a commit that adds `Engine` back to the API, changes `StoreBuilder` back to `EngineBuilder`, and `Host` now takes the engine to use rather than a `Store`.

I've also added the examples to CI to build and run them.

PR #37, opened 2020-09-05
Use nullable reference types to annotate nullability

SirJosh3917:
With C# 8 brings NRT (Nullable Reference Types) which enforce nullability at compile time of reference types. This PR updates the signatures of method headers, properties, and other relevant data types to use NRTs where applicable to more clearly inform developers that certain return values may be null when they are not.

Perhaps the possibility of removing many of the manual null parameter checks could be accompanied for in this PR. Removing the nullability checks in methods yields (minor) performance gains, and removes lots of boilerplate from the codebase.


peterhuene (approved):
Sorry for the delay in reviewing this.  This looks great!  Thank you for taking the time to add these annotations.

PR #38, opened 2020-09-06
Enhance array-based methods with Span<T> API

SirJosh3917:
In 2018, Span<T> and its friends were added to C#. This allowed developers to wrap various kinds of contiguous memory buffers (arrays, pointers, others) into a single data type. Its ability to slice allowed developers to very ergonomically prevent copying data, in a .NET way. This PR overhauls the existing public API to make use of Span<T> and friends.

For `Interop.InvokeCallback`, I would like feedback on why the object array was being allocated outside the invocation. This PR may need more inspection on that front, but for me on Windows all the tests pass.

In addition, DLL imports from the Interop class were modified from arrays to pointers, as it's a more "universal" datatype and understood the exact same. You can convert arrays into pointers easily, but you can't convert stackalloc'd or other pointers into an array easily. I would like feedback on this change as well.

### DIFF
--- a/src/Externs/ExternFunction.cs
+++ b/src/Externs/ExternFunction.cs
@@ -44,6 +44,21 @@ namespace Wasmtime.Externs
             return Function.Invoke(_func, Parameters, Results, arguments);
         }
 
+        // TODO: remove overload when https://github.com/dotnet/csharplang/issues/1757 is resolved
+        /// <summary>
+        /// Invokes the WebAssembly function.
+        /// </summary>
+        /// <param name="arguments">The array of arguments to pass to the function.</param>
+        /// <returns>
+        ///   Returns null if the function has no return value.
+        ///   Returns the value if the function returns a single value.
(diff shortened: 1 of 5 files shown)


peterhuene:
Apologies for the delay in reviewing this one as well.  We'll need a rebase following the merge of the NRT annotations.

SirJosh3917:
Alright, rebased but this test fails and I don't know if I should be concerned or not:

image

peterhuene:
@SirJosh3917 apologies for the delayed response.  The CI failure was a result of an upstream change in Wasmtime.  I am fixing it now and will rebase your PR and merge once green.  No further action is required on your part.

peterhuene (approved):

peterhuene:
@SirJosh3917 thanks for the work on this!  The changes look good and I'll cut a 0.20.0 release of the NuGet package with these changes (and update Wasmtime to latest, which is currently 0.20.0).

PR #41, opened 2020-10-30
Fix CI

peterhuene:


tests/InvalidModuleTests.cs
```diff
@@ -16,7 +16,7 @@ public void ItThrowsWithErrorMessageForInvalidModules()
             action
                 .Should()
                 .Throw<WasmtimeException>()
-                .WithMessage("WebAssembly module 'invalid' is not valid: Unexpected EOF (at offset 0)");
+                .WithMessage("WebAssembly module 'invalid' is not valid: WebAssembly failed to compile*");
```
yurydelendik, line 19: has the trailing `*` any meaning?

yurydelendik (approved):

tests/InvalidModuleTests.cs
peterhuene: It's a wildcard to let the assert match the rest of the message, which may contain a callstack (on platforms that can walk through the foreign frames correctly).

peterhuene: FYI, Fluent doesn't use real regexes here, just some limited wildcard tokens.

Issue #45, opened 2020-12-03
Implement support for module linking


peterhuene:
See the C API changes [here](https://github.com/bytecodealliance/wasmtime/pull/2472#pullrequestreview-544481017).

PR #48, opened 2021-01-14
Update Wasmtime version to 0.21.0.

peterhuene:


peterhuene:
CI will fail for this as it's testing against upstream's breaking changes in main.

I've tested this locally and tests pass, so not going to block on this.  When this is merged, I'll push 0.21.0 to NuGet.

Following that, I'll implement #45 and bump the version to 0.22.0.

Once 0.22.0 is published, I'll submit a fix for the the breaking changes in main.

PR #49, opened 2021-01-16
Implement module linking. 

### LINKED ISSUE #40: No way to get access to exported memory outside of a Host callback
Inside a host defined callback we can do this:
```
host.DefineFunction("ns", "someHostFunction", (Caller caller, int data_ptr, int data_len) => {
       var mem = caller.GetMemory("memory"); 
}
```
However there is no exposed API for getting access to a memory via the instance directly like you can do within rust:
```
let memory = instance.get_memory("memory").ok_or(anyhow::format_err!("failed to find `memory` export"))?;	
```

### LINKED ISSUE #45: Implement support for module linking

peterhuene:
This PR implements the changes to support the module linking proposal.

Modules and instances are now supported import types for module instantiation.

Modules can now also be instantiated via `Module.Instantiate` and
`ModuleExtern.Instantiate`.

There is a breaking change to the API:

* `Instance.Externs` has been removed in favor of individual properties that
  forward to the inner `Externs` instance.  Thus `Instance.Externs.Functions`
  becomes `Instance.Functions`, etc.

Additionally, the examples and the test project have been migrated to .NET 5.

Closes #45.
Fixes #40.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,19 +53,10 @@ jobs:
       if: matrix.config == 'release'
       run: dotnet run -c ${{ matrix.config }} --project benchmarks/simple/simple.csproj
     - name: Run examples
+      env:
+        EXAMPLES: externref funcref global hello linking memory table
       run: |
-        cd examples/hello
-        dotnet run
-        cd ../global
-        dotnet run
-        cd ../memory
-        dotnet run
-        cd ../table
-        dotnet run
-        cd ../externref
-        dotnet run
-        cd ../funcref
-        dotnet run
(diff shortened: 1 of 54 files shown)


peterhuene:
CI is expected to fail for this PR as there's a breaking change upstream in Wasmtime which is what the CI is testing against. 

Tests pass locally so I won't block this PR on CI.

The CI will be fixed once the `0.22.0` NuGet package is published.

peterhuene:
Merging as tests pass locally (see comment above).

PR #58, opened 2021-04-20
Fix tests.

peterhuene:
Fix some of the tests given changes to upstream Wasmtime.

Imports need to be resolved prior to matching on the types now, so the tests
are updated to define the missing imports.

Additionally, the expected error message changed slightly.


peterhuene:
Looks like there's still a funcref test failure on Windows; I'm going to merge this as it at least gets other tests green and figure out what's still broken.

PR #62, opened 2021-06-14
Reimplement the Wasmtime API for .NET.

### LINKED ISSUE #53: Race in SafeHandle finalizers leads to undefined behavior
As store-based objects in the C API need to used from a single thread, the finalizers of the various `SafeHandle`-derived types that call `wasm_<type>_delete` introduce a race on the underlying thread-unsafe reference count of the Wasmtime `Store`.

This may lead to undefined behavior when one thread is modifying the store's reference count and the finalizer thread is decrementing the reference count: sometimes leaking the `Store` (and the associated objects) or prematurely deallocating the `Store` while still in use.

A workaround may be needed in the .NET API to account for this limitation of the C API.

**Users may also work-around this problem by always calling `Dispose` on any `IDispoable` type in the Wasmtime .NET API.**

Ideally, Wasmtime would allow calls to `wasm_<type>_delete` from a different thread, provided the given object is no longer in use by any other thread (guaranteed to be the case for a `SafeHandle` finalizer in the .NET API).

peterhuene:
This PR reimplements the Wasmtime API for .NET in terms of the latest
upstream changes.

Recently the Wasmtime Rust API underwent a refactoring that made Wasmtime safer
to use in multithreaded async executors.

As a result of that change, a "store context" was needed to pass through to all
store-related types. Additionally, the store-related types no longer need to
implement IDisposable because the owning store now properly handles cleanup.
This fixes a race that previously may have occurred when .NET users forgot to
call `Dispose` and cleanup was done from a finalizer.

As this was already a breaking change, I decided to remove `Host` in favor of a
`Linker` type to better encapsulate the underlying Wasmtime API. `Host` existed
before the linker implementation in Wasmtime and I have been meaning to remove
it for a while now.

Another big change is that `Instance` no longer implements `DynamicObject`.
Users should now call `Instance.GetFunction`, etc. for retrieving the exports
of an instance.

Examples, documentation, and tests were all updated to the new API.

Fixes #53.


src/Store.cs
```diff
@@ -1,13 +1,165 @@
...
-using System.IO;
-using System.Text;
-using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using Microsoft.Win32.SafeHandles;
 
 namespace Wasmtime
 {
+    /// Represents a Wasmtime store context.
+    public readonly ref struct StoreContext
```
alexcrichton, line 8: FWIW I've avoided exposing this as a first-class type in other bindings because it's in general not safe to use (since it can be used outside of the lifetime of the `Store` itself). I've been adding these methods to `Store` and then having any method that takes a `Store` take anything that you can get a context from, but acquiring the context is just an internal operation.

src/MemoryBase.cs
```diff
@@ -1,56 +1,67 @@
...
 using System.Text;
 using System.Buffers.Binary;
+using System.Runtime.InteropServices;
 
 namespace Wasmtime
 {
+    /// <summary>
+    /// The base class for memories.
+    /// </summary>
     public abstract class MemoryBase
```
alexcrichton, line 11: Out of curiosity, how come there's a split between this and `Memory`?

src/MemoryBase.cs
```diff
@@ -63,189 +74,225 @@ public string ReadNullTerminatedString(int address)
...
 
         /// <summary>
         /// Grows the memory by the specified number of pages.
         /// </summary>
+        /// <param name="context">The store context for the memory.</param>
         /// <param name="delta">The number of WebAssembly pages to grow the memory by.</param>
-        /// <returns>Returns true if the memory grew or false if it would exceed the maximum size.</returns>
-        public bool Grow(uint delta)
+        /// <returns>Returns the previous size of the Webassembly memory, in pages.</returns>
+        public uint Grow(StoreContext context, uint delta)
```
alexcrichton, line 271: This may want to mention that this invalidates all previous spans?

src/Linker.cs
```diff
@@ -0,0 +1,382 @@
...
+        /// <param name="module">The module to instantiate.</param>
+        /// <returns>Returns the new instance.</returns>
+        public Instance Instantiate(StoreContext context, Module module)
+        {
+            if (module is null)
+            {
+                throw new ArgumentNullException(nameof(module));
+            }
+
+            var error = Native.wasmtime_linker_instantiate(handle, context.handle, module.NativeHandle, out var instance, out var trap);
```
alexcrichton, line 141: This is a point where wasm code runs so in other bindings I typically expect it to have roughly the same "guards" as `Func::call` (in w/e form that has). One thing this might be missing is handling of the exception propagation? 

(I'm just assuming that all C# exceptions are caught, converted to a trap, and then when the trap is received on the other side we realize that  the C# exception should be re-raised. Assuming that unwinding via C# through Rust code is unsound)

src/Linker.cs
```diff
@@ -0,0 +1,382 @@
...
+            {
+                throw new ArgumentNullException(nameof(module));
+            }
+
+            unsafe
+            {
+                var nameBytes = Encoding.UTF8.GetBytes(module.Name);
+                fixed (byte* namePtr = nameBytes)
+                {
+                    var error = Native.wasmtime_linker_module(handle, context.handle, namePtr, (UIntPtr)nameBytes.Length, module.NativeHandle);
```
alexcrichton, line 172: Technically I think this can do instantiation like above as well

src/Instance.cs
```diff
@@ -1,136 +1,214 @@
...
+                {
+                    var external = imports[i] as IExternal;
+                    if (external is null)
+                    {
+                        throw new ArgumentException($"Objects of type `{imports[i].GetType().ToString()}` cannot be imported.");
+                    }
+                    externs[i] = external.AsExtern();
+                }
+
+                var error = Native.wasmtime_instance_new(context.handle, module.NativeHandle, externs, (UIntPtr)imports.Length, out this.instance, out var trap);
```
alexcrichton, line 44: I had more of a comment in `Linker` below, but this is another point where wasm code is executed

src/Function.cs
```diff
@@ -821,119 +633,111 @@ private static bool IsTuple(Type type)
...
+                            results[i] = Value.FromObject(tuple[i], resultKinds[i]);
                         }
                     }
                 }
                 return IntPtr.Zero;
             }
             catch (Exception ex)
             {
-                var bytes = Encoding.UTF8.GetBytes(ex.Message + "\0" /* exception messages need a null */);
+                var bytes = Encoding.UTF8.GetBytes(ex.Message);
```
alexcrichton, line 677: Oh my comments below may be wrong in reference to this.

FWIW for other language bindings what I tried to do with native-language exceptions was to treat them similarly to Rust panics where we propagate them over wasm. The implementation of all the languages is to indeed return a trap here, but upon catching the trap on the other side it re-raises the exception object which was stored somewhere else.

For example in Python [this wraps entry into wasm](https://github.com/bytecodealliance/wasmtime-py/blob/64f5145a0ed5a5d8afe2203f46a25ae9303c9713/wasmtime/_func.py#L255-L266) where [this catches all exceptions in host callbacks](https://github.com/bytecodealliance/wasmtime-py/blob/64f5145a0ed5a5d8afe2203f46a25ae9303c9713/wasmtime/_func.py#L208-L214). I'm not sure if it makes sense for C# to do the same thing, but it might help debugging in some cases where native-language bugs don't get accidentally swallowed into traps?

src/Function.cs
```diff
@@ -682,94 +500,88 @@ internal Function(Interop.StoreHandle store, Delegate callback, bool hasReturn)
...
+                Native.WasmtimeFuncCallback? func = null;
+                if (hasCaller)
+                {
+                    func = (env, caller, args, nargs, results, nresults) =>
+                        InvokeCallback(callback, new Caller(caller), true, args, (int)nargs, results, (int)nresults, Results);
+                }
+                else
+                {
+                    func = (env, caller, args, nargs, results, nresults) =>
+                        InvokeCallback(callback, new Caller(caller), false, args, (int)nargs, results, (int)nresults, Results);
```
alexcrichton, line 521: In other bindings I have a `finally` block which nulls out the interior pointer in `Caller` to signify that it's no longer valid to use (ideally preventing accidental bugs where the `Caller` object is stored to outlive the call and then it's used again). 

(e.g. something [like this](https://github.com/bytecodealliance/wasmtime-py/blob/64f5145a0ed5a5d8afe2203f46a25ae9303c9713/wasmtime/_func.py#L215-L216))

src/EngineBuilder.cs
```diff
@@ -54,7 +73,45 @@ public class EngineBuilder
         /// <returns>Returns the current builder.</returns>
         public EngineBuilder WithDebugInfo(bool enable)
```
alexcrichton, line 74: I think this is basically serving the position of `Config`, right? If that's true perhaps for consistency with other bindings this could be called `Config`?

I was also curious as to why you stored each configuration field's value in this struct, and then have a "build" which actually feeds them all into a `wasm_config_t`. Alternatively the `EngineBuilder` could just hold on to the `wasm_config_t`, right?

alexcrichton (commented):
👍 

Seems pretty reasonable to me! I don't really know a ton about C# so I was mostly looking for high-level things compared to other language bindings

src/MemoryBase.cs
peterhuene: Before `Memory` was encapsulating an owned `wasm_memory_t` and `CallerMemory` was encapsulating an owned `wasm_extern_t`; `MemoryBase` was just a way to share the implementation between the two.

With the refactoring there's no reason for `CallerMemory` any longer, so I can remove this and consolidate `MemoryBase` with `Memory`.

src/Linker.cs
peterhuene: Any calls from wasm to .NET will catch .NET exceptions at the entry site and return a trap back as the result of the host function callback.

There shouldn't be any (catchable) .NET exceptions propagating out from a call to `wasmtime_linker_instantiate`; we check that the returned trap is non-zero and then throw the trap exception.

src/Function.cs
peterhuene: I could certainly add something similar to .NET, but I think that can be a follow-up PR.

Realistically speaking, if I know an exception is coming from .NET it's trivial to break a debugger on a "first chance" exception to find exactly where the host call is throwing.

However, it's nigh impossible to debug via a log file without having the original stack trace as part of `TrapException`, which would require us to have a side channel to store the original exception object.

src/Function.cs
peterhuene: Changing `Caller` to `ref struct` should prevent storage of it beyond this stack frame.  I can make that fix.

src/MemoryBase.cs
peterhuene: I'll update the comment for that.

src/Store.cs
peterhuene: `ref struct` forces the compiler to ensure that this type can only be allocated on a stack frame, so it will make it much harder for a store context to outlive the store (obviously not impossible as you can explicitly dispose a `Store` with the context still stored in the same frame).

I'll see about making `StoreContext` an internal type.

src/EngineBuilder.cs
peterhuene: I think the original intention might have been to only throw from `Build` and not the various setter methods, but there's no reason for that really; it's certainly not worth the added complexity so I can remove the fields.

Regarding the name, I'm fine with renaming this to `Config` for consistency as now would be the time to do it.

src/Function.cs
peterhuene: Although I just remember why I can't do that as we invoke the .NET callback via reflection currently, so that can't happen if it's `ref struct`.

I'll go with invalidating the interior pointer and throwing an exception when accessed again.

src/EngineBuilder.cs
peterhuene: Actually, I believe the actual intention for why `EngineBuilder` was done this way was to solve the ownership transfer problem given that `wasm_engine_new_with_config`, sadly, takes ownership of the passed in config.

Having the config built entirely in the `Build` method prevented the `EngineBuilder` from having to implement `IDisposable`.

peterhuene:
@alexcrichton I've pushed up review feedback for what I feel needs to be addressed as part of this PR.

src/Linker.cs
peterhuene: Opened issue #63 for this.

src/Linker.cs
peterhuene: Opened issue #63 for this.

src/Instance.cs
peterhuene: Opened issue #63 for this.

src/Function.cs
peterhuene: Opened issue #63 for this.

alexcrichton (approved):

PR #94, opened 2022-02-24
Upgrade to Dotnet 6.

alexhiggins732:
The current build targets [ .Net Standard 2.1](https://github.com/bytecodealliance/wasmtime-dotnet/blob/19b3bccf841805481d2b657b0eb1b83a0a01b11f/src/Wasmtime.csproj#L4) in the main library and [.Net 5.0](https://github.com/bytecodealliance/wasmtime-dotnet/blob/19b3bccf841805481d2b657b0eb1b83a0a01b11f/tests/Wasmtime.Tests.csproj#L4) in the unit test library.

Microsoft announced [September 15th, 2020]( https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/) that .Net 5.0 replaced  .Net Standard 2.1 and with .Net 6.0, which is now the current LTS channel, replacing .Net 5.0 for which support officially ends in less than 3 months on [May 8, 2022](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core).

Additionally, the current build configuration also has a dependency on at least 2 .Net 4.X framework DLL's - `Microsoft.Csharp (4.6.0)` and `System.Configuration.ConfigurationManager(4.4.0)` 

image

Upgrading both the main library and the unit tests to  .Net 6, and upgrading packages to their latest versions removes the legacy 4.x DLL dependencies and moves Wasmtime-dotnet onto the current LTS version of dotnet.

Additionally, the Unit tests for `Wasmtime.ValueType` were updated to perform `null` checks to allow compiling on .Net 6.

### DIFF
--- a/src/Value.cs
+++ b/src/Value.cs
@@ -210,25 +210,25 @@ namespace Wasmtime
                 {
                     case ValueKind.Int32:
                         if (o is null)
-                            throw new WasmtimeException($"The value `{o ?? "null"}` is not valid for WebAssembly type {kind}.");
+                            throw new WasmtimeException($"The value 'null' is not valid for WebAssembly type {kind}.");
                         value.of.i32 = (int)Convert.ChangeType(o, TypeCode.Int32);
                         break;
 
                     case ValueKind.Int64:
                         if (o is null)
-                            throw new WasmtimeException($"The value `{o ?? "null"}` is not valid for WebAssembly type {kind}.");
+                            throw new WasmtimeException($"The value 'null' is not valid for WebAssembly type {kind}.");
                         value.of.i64 = (long)Convert.ChangeType(o, TypeCode.Int64);
                         break;
 
                     case ValueKind.Float32:
                         if (o is null)
-                            throw new WasmtimeException($"The value `{o ?? "null"}` is not valid for WebAssembly type {kind}.");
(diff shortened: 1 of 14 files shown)


src/Value.cs
```diff
@@ -209,18 +209,26 @@ public static Value FromObject(object? o, ValueKind kind)
                 switch (kind)
                 {
                     case ValueKind.Int32:
+                        if (o is null)
+                            throw new WasmtimeException($"The value `{o ?? "null"}` is not valid for WebAssembly type {kind}.");
```
peterhuene, line 213: If `o` is `null`, do we need the `{o ?? "null"}` in the string interpolation? 

Perhaps simply `A null value is not valid for WebAssembly type {kind}.`?

Same comment here as for the rest of the string interpolations in this function.

peterhuene (approved):
This looks good, thank you for taking the type to update this repo to `net6.0`.

Just the one comment.

src/Value.cs
peterhuene: Looks like the tests need updating as well.

alexhiggins732:
I conformed the exception message to match what the unit tests expecte:

```
$"The value `null` is not valid for WebAssembly type {kind}."
```

peterhuene:
@alexhiggins732 thanks for fixing this!

Issue #106, opened 2022-04-16
Manually consume fuel from store


my0n:
Is it possible to manually consume fuel from the store? Looks like adding fuel and getting the current amount of consumed fuel is supported, but I wanted to put an extra cost to some of the functions that I'm calling.

I'm very new to wasm stuff in general, but it looks like this should be possible using [Store::consume_fuel](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html#method.consume_fuel) - calling that function seems straightforward so I may be able to take a swing at this if you'd like

peterhuene:
Hi @my0n. Indeed, `Store::consume_fuel` would be the way to do this. It is exposed in Wasmtime's C API as `wasmtime_context_consume_fuel`, so it should be usable from the .NET API with a P/Invoke added.

It would be wonderful if you could take a stab at adding support for it and putting up a PR; I'm happy to help if you run into any problems!

my0n:
Opened up a PR! Also thanks for the library, it's very clean and easy to use

peterhuene:
Implemented in #107.

PR #107, opened 2022-04-19
Add fuel consumption tests and consume fuel

my0n:
Here's a PR for https://github.com/bytecodealliance/wasmtime-dotnet/issues/106

* Added tests for existing Store fuel functionality, specifically AddFuel and GetConsumedFuel.
* Added a virtual method GetEngineConfig to the ModuleFixture class to add customized Engine configs.
* Added a pinvoke call to wasmtime_context_consume_fuel and wired that up through the Store.
* Added tests for the above.
* Added an example for consuming fuel.

### DIFF
--- a/Examples.sln
+++ b/Examples.sln
@@ -17,6 +17,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "table", "examples\table\tab
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Wasmtime", "src\Wasmtime.csproj", "{82B01E4A-1CAD-44BB-B923-11FA37173BD7}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "consumefuel", "examples\consumefuel\consumefuel.csproj", "{F79FAA27-3CA6-4CC3-BB50-CE27191770F1}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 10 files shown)


Examples.sln
```diff
@@ -39,10 +41,6 @@ Global
 		{D3D124D2-AE6F-49F5-81F4-8FE41451856B}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{D3D124D2-AE6F-49F5-81F4-8FE41451856B}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{D3D124D2-AE6F-49F5-81F4-8FE41451856B}.Release|Any CPU.Build.0 = Release|Any CPU
-		{5C690486-153E-4221-A3FD-BE6B8B4FE579}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{5C690486-153E-4221-A3FD-BE6B8B4FE579}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{5C690486-153E-4221-A3FD-BE6B8B4FE579}.Release|Any CPU.ActiveCfg = Release|Any CPU
-		{5C690486-153E-4221-A3FD-BE6B8B4FE579}.Release|Any CPU.Build.0 = Release|Any CPU
```
my0n, line 45: Visual Studio automatically made this change - it looks like this guid was for the `linking` example project that was removed in 8984b08d9408b1857bcaf0b9d65f489e3fda8335.

examples/consumefuel/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+            using var store = new Store(engine);
+
+            linker.Define(
+                "",
+                "expensive",
+                Function.FromCallback(store, (Caller caller) =>
+                {
+                    var remaining = store.ConsumeFuel(1000UL);
+                    Console.WriteLine($"Called an expensive function which consumed 1000 fuel. {remaining} units of fuel remaining.");
+                }
```
my0n, line 23: My main goal for making the example project was to make sure I'm not calling ConsumeFuel in a weird way here - I suspect this is how users (myself included) are going to want to call the method but lmk if doing it like this can e.g. cause multithreading issues.

examples/consumefuel/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+            using var module = Module.FromTextFile(engine, "consumefuel.wat");
+            using var linker = new Linker(engine);
+            using var store = new Store(engine);
+
+            linker.Define(
+                "",
+                "expensive",
+                Function.FromCallback(store, (Caller caller) =>
+                {
+                    var remaining = store.ConsumeFuel(1000UL);
```
peterhuene, line 21: I think we'd want to use `caller` rather than capturing `store`, if possible.

However `Caller` doesn't expose the fuel methods like `Store` does. Do you mind adding `AddFuel`, `ConsumeFuel`, and `GetConsumedFuel` methods to `Caller` (should be identical to those in `Store`)?

That way we can just do:

```c#
var remaining = caller.ConsumeFuel(1000UL);
```

And the callback has no captured variables.

examples/consumefuel/Program.cs
peterhuene: I don't foresee any problems given the API's inherent multi-threading issues; lots of types are safe to send between threads (i.e. have no thread-affinity) but not to use from multiple threads simultaneously.

peterhuene (commented):
This looks really great! Thanks for putting this up 👍.

The changes all look good, just the one comment regarding implementing some fuel-related methods for `Caller`.

examples/consumefuel/Program.cs
my0n: Agreed - much better. Thanks!

also snuck in one last change - I added the example to CI (didn't see the build step that runs the examples until now)

peterhuene (approved):
Perfect, thanks again for taking the time to implement this!

PR #111, opened 2022-05-14
Added an ability to get function from Caller

mralexes:
Allows getting functions from Caller in the scope of imported function.
```csharp
         linker.Define("env", "computeSha256", Function.FromCallback(store,
                (Caller caller, int dataPtr) =>
                {
                    var memory = caller.GetMemory("memory");
                    var data = ReadBytes(dataPtr, memory, caller);
                    var hash = SHA256.Create().ComputeHash(data);

                    var allocFunc = caller.GetFunction("alloc");
                    var hashPtr = (int) allocFunc.Invoke(store, hash.Length);

                    hash.CopyTo(memory.GetSpan(store).Slice(hashPtr, hash.Length));
                }));
```

### DIFF
--- a/src/Caller.cs
+++ b/src/Caller.cs
@@ -48,6 +48,35 @@ namespace Wasmtime
             }
         }
 
+        /// <summary>
+        /// Gets an exported function of the caller by the given name.
+        /// </summary>
+        /// <param name="name">The name of the exported function.</param>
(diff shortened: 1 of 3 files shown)


peterhuene (commented):
Thanks for adding this! This seems like a reasonable change to me.

Would you mind adding a simple test that gets a function from the caller and calls it, perhaps?

peterhuene (approved):

Issue #113, opened 2022-05-31
Proposal: Reduce Boxing


martindevans:
I'm currently looking into using wasmtime-dotnet in a Unity game. Generally in Unity you want to reduce allocations as much as possible to prevent any GC stuttering down the line. As it stands any time you invoke a wasm function there is some unavoidable boxing of primitive arguments because the arguments are passed as `params object[]` (the `ReadOnlySpan<object>` variant of the call avoids the allocation of the params array, but not the allocation caused by boxing all of the individual arguments).

I can see two possible workaround for this.

## Generic Invoke

Define a large number of overloads for Invoke with varying numbers of generic arguments:

```csharp
public object? Invoke<TA>(IStore store, TA arg1);
public object? Invoke<TA, TB>(IStore store, TA arg1, TB arg2);
public object? Invoke<TA, TB, TC>(IStore store, TA arg1, TB arg2, TC arg3);
// etc up to as many arguments as you want to support
```

For the sake of sanity, writing this would probably require code generation!

## Struct Boxing

An alternative option would be to define a new struct which works in basically the same way as the `ValueUnion` struct (let's call it `ArgBox`) and to have implicit casts from the various supported types into this new struct. Then there just need to be two new methods defined:

```csharp
public object? Invoke(IStore store, params ArgBox?[] args);
public object? Invoke(IStore store, ReadOnlySpan<ArgBox?> args);
```

This is essentially just the same boxing operation that is already going on, but without the need for silent heap allocations.

## Interest?

Does this proposal sound reasonable and would you be interested in merging a PR that contained something like this (and if so, which one)? I'm willing to write it if so.

peterhuene:
Hi @martindevans. I think the proposal is absolutely reasonable as this was definitely one of the areas I wanted to improve but never got around to doing so.

Between the two options, I'd favor a value type with implicit conversions rather than further overloads to `Invoke`.

Very much willing to review and merge anything that improves this!

martindevans:
Resolved by #114

PR #114, opened 2022-06-01
ValueBox

martindevans:
See proposal #113.

To reduce allocations for boxing on every call this PR introduces a new `ValueBox` struct which wraps up arguments without requiring any heap allocations. Currently this is only used for function arguments, but in principle it could be used anywhere else plain old `object` references are passed around (e.g. function returns values).

## Drawbacks

1. `ValueBox` is _mostly_ invisible to the end user since things convert to it with an implict conversion. However, C# does not support user defined conversions from `object` so that has to be done with an explicit conversion method - i.e. passing `my_ref_object` becomes `ValueBox.AsBox(my_ref_object)`. I've added an implict conversion for strings, since that's likely to be a fairly common case. Unfortunately I don't see a better way to handle this situation for any other references types.

2. I had to remove the old `params object[]` method, since it confliced with the new `params ValueBox[]` method, which is a breaking API change. If you'd rather not much such a big change to the external API I can instead remove `params ValueBox[]` and require that users call it explicitly with a `ReadOnlySpan<ValueBox>` instead.

3. `params ValueBox[]` still incurs a single allocation, which can be worked around by explicitly passing in a `ReadOnlySpan<ArgBox>`, but that's not very ergonomic.

4. I haven't done anything with return types at all, they still get boxed. It's not obvious how to fix them, since the caller does not supply the return type.

## Potential Improvements

We could potentially workaround a lot of the drawbacks above by introducing a new set of generic `GetFunction` overloads which return `Func<>` or `Action<>` objects. This would have quite a few benefits:
 - Compatibility of types could be checked when `GetFunction` is called, instead of when `Invoke` is called. Potentially reducing per-call overhead.
 - Fixed number of arguments, so ergonomics issues with params array/span disappear.
 - Conversion to `ValueBox` can be done inside the `Func`/`Action` which is returned, making it invisible to the end user (`ValueBox.AsBox` doesn't need to be called externally)
 - Return type unboxing can be handled easily in the same way.
 - Fits better into a normal C# workflow - wasm functions can be used just like any other `Func`/`Action` object.

If you're interested I can put together another PR which does this (don't merge this PR if so).

### DIFF
--- a/src/Function.cs
+++ b/src/Function.cs
@@ -14,6 +14,7 @@ namespace Wasmtime
     /// </summary>
     public class Function : IExternal
     {
+        #region FromCallback
         /// <summary>
         /// Creates a function given a callback.
         /// </summary>
@@ -523,6 +524,7 @@ namespace Wasmtime
 
             return new Function(store.Context, callback, true);
         }
+        #endregion
 
         /// <summary>
         /// The parameters of the WebAssembly function.
@@ -545,6 +547,25 @@ namespace Wasmtime
         public static Function Null => _null;
 
         /// <summary>
+        /// Invokes the Wasmtime function with no arguments.
+        /// </summary>
+        /// <param name="store">The store that owns this function.</param>
+        /// <returns>
+        ///   Returns null if the function has no return value.
+        ///   Returns the value if the function returns a single value.
+        ///   Returns an array of values if the function returns more than one value.
+        /// </returns>
+        public object? Invoke(IStore store)
+        {
+            if (store is null)
+            {
+                throw new ArgumentNullException(nameof(store));
+            }
+
+            return Invoke(store, new ReadOnlySpan<ValueBox>());
+        }
+
+        /// <summary>
         /// Invokes the Wasmtime function.
         /// </summary>
         /// <param name="store">The store that owns this function.</param>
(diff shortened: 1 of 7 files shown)


peterhuene:
@martindevans if you rebase your PR on main, that should resolve the test failures.

tests/ExternRefTests.cs
```diff
@@ -49,7 +49,7 @@ public void ItHandlesNullReferences()
             var nullref = instance.GetFunction(Store, "nullref");
             inout.Should().NotBeNull();
 
-            (inout.Invoke(Store, null)).Should().BeNull();
+            (inout.Invoke(Store, ValueBox.AsBox((object?)null))).Should().BeNull();
```
peterhuene, line 52: As these are tests, I think we can just omit the `?` here to remove the warning.

peterhuene (commented):
These changes look good to me, just a few comments to address.

Regarding this approach and its drawbacks, I think it's beneficial enough to justify any breaking changes around passing arguments existing users might encounter.

Obviously some sort of function wrapping where we can return a callable object where the argument type checks are elided for the actual call would be most ideal. If you want to take a stab at that approach, I'd be happy to review it in lieu of or in addition to merging this.

tests/ExternRefTests.cs
martindevans: This actually lead me to noticing that `AsBox` should accept `null` anyway.

src/Function.cs
```diff
@@ -554,80 +575,74 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// <param name="arguments">The arguments to pass to the function, wrapped in `ArgBox`</param>
+        /// <returns>
+        ///   Returns null if the function has no return value.
+        ///   Returns the value if the function returns a single value.
+        ///   Returns an array of values if the function returns more than one value.
+        /// </returns>
+        public object? Invoke(IStore store, ReadOnlySpan<ValueBox> arguments)
         {
             if (IsNull)
-            {
```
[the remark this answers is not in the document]
martindevans: I've added back in all of the missing braces. I'll let you mark this as resolved if it's to your taste :)

peterhuene (approved):
👍 looks good to merge.

Would you like this to go in now or would you like to instead wait for taking the alternative approach of returning `Func`-like wrappers that elide the parameter / result conversions and type checking?

martindevans:
We may as well leave this PR for a couple of days, just in case I need to make any major changes to it to support the Func-wrapper approach.

martindevans:
So far I didn't have to make any changes to this for my #116 draft PR, so I think it's probably safe to merge it as is. I don't mind if you'd rather wait and do it all in one go though :)

martindevans:
I've swapped out the representation of `v128` from a `byte[]` (requires an allocation, size must be runtime checked) to `Vector128<byte>` (no allocations, size is part of type) as proposed in #120. This change is required for #116 to move forward.

peterhuene:
I agree, that is much nicer!

Sorry, I've been out sick this week, so I haven't been responding to GitHub notifications.

Please let me know if you want me to review this again and merge it if it's needed for the other PR.

martindevans:
That's no problem, I hope you're feeling better now :)

Please go ahead and review this and merge it if you're happy with it. The other PR builds on top of both changes this introduces.

src/V128.cs
```diff
@@ -0,0 +1,99 @@
...
+{
+    /// <summary>
+    /// A 128 bit value
+    /// </summary>
+    public struct V128
+    {
+        /// <summary>
+        /// Get a V128 with all bits set to 1
+        /// </summary>
+        public static readonly V128 AllBitsSet = new V128(
```
peterhuene, line 14: Should we call this `MaxValue` instead?

src/V128.cs
```diff
@@ -0,0 +1,99 @@
...
+        private unsafe fixed byte Value[16];
+
+        /// <summary>
+        /// Construct a new V128 from a 16 element byte span
+        /// </summary>
+        /// <param name="bytes"></param>
+        /// <exception cref="ArgumentException"></exception>
+        public V128(ReadOnlySpan<byte> bytes)
+        {
+            if (bytes.Length != 16)
```
peterhuene, line 30: Nit: could we use braces around  this conditional's body?

src/V128.cs
```diff
@@ -0,0 +1,99 @@
...
+                throw new ArgumentException("Must supply exactly 16 bytes to construct V128");
+
+            unsafe
+            {
+                fixed (byte* value = Value)
```
peterhuene, line 35: Nit: could we use braces around this `fixed` statement's body?

src/V128.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            fixed (byte* value = Value)
+            {
+                Unsafe.CopyBlock(value, src, 16);
+            }
+        }
+
+        internal unsafe void CopyTo(byte* dest)
+        {
+            for (int i = 0; i < 16; i++)
```
peterhuene, line 92: Perhaps we should call `Buffer.MemoryCopy` (or `Unsafe.CopyBlock`; I can't say which would be a better optimization without looking into it) rather than a loop?

src/Value.cs
```diff
@@ -328,24 +324,8 @@ public static Value FromObject(object? o, ValueKind kind)
...
-                            of.v128[8],
-                            of.v128[9],
-                            of.v128[10],
-                            of.v128[11],
-                            of.v128[12],
-                            of.v128[13],
-                            of.v128[14],
-                            of.v128[15]
-                        );
+                        fixed (byte* v128 = of.v128)
```
peterhuene, line 327: Nit: braces around `fixed` statement.

src/V128.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                fixed (byte* value = Value)
+                    bytes.CopyTo(new Span<byte>(value, 16));
+            }
+        }
+
+        /// <summary>
+        /// Construct a new V128 from 16 bytes
+        /// </summary>
+        /// <param name="e0"></param>
```
peterhuene, line 43: Nit: perhaps these should be `b0` to `b15`? Also, for completion, could we fill in these param comments, even if it's a rather meaningless "The first byte value", "The second byte value", etc.?

peterhuene (commented):
Just a comment or two and a few nits.  Looking really great!

src/V128.cs
peterhuene: And perhaps also update the doc comment to indicate it is the maximum value of the `v128`?

peterhuene: Although, the more I think about this, a `v128` is intended to be a vector of bits more akin to a SIMD register, so calling this the "maximum" value doesn't seem right.

I'm fine to leave this as-is.

martindevans: I was thinking of it more in terms of a vector of 16 bytes (or 4 ints/floats etc), rather than a single 128 bit value, so I personally think the current name makes more sense. Final decision is yours of course.

src/V128.cs
martindevans: That's a much better name. Honestly not sure what I was htinking of with `e`...

src/V128.cs
martindevans: I rewrote this to use `Span<byte>.CopyTo`, should be the best option I think.

martindevans:
I've addressed all of that feedback

PR #116, opened 2022-06-05
Type safe functions

martindevans:
## New public methods on `Instance`
 - `public Function? GetFunction(IStore store, string name, Type? returnType, params Type[] parameterTypes)`
   - Gets a `Function` only if it matches the specified type signature
 - `public Action<TA>? GetAction<TA>(IStore store, string name)`
   - Gets and wraps a Function which returns nothing
 - `public Func<TR>? GetFunction<TR>(IStore store, string name)`
   - Gets and wraps a Function which takes no arguments and returns something
 - `public Func<TA, TR>? GetFunction<TA, TR>(IStore store, string name)`
   - Gets and wraps a Function which takes 1 argument and returns something
 - `public Func<TA, TB, TR>? GetFunction<TA, TB, TR>(IStore store, string name)`
   - Gets and wraps a Function which takes 2 arguments and returns something

In all of these cases multiple return types are handled as tuples. That looks like this in use: `instance.GetFunction<int, int, (int, int)>("swap")`

## New Internals

 - `Function` has an internal `CheckTypeSignature` method which simply checks if a given type signature matches this function. Could be useful to make public?
 - `Function` has various internal `WrapFunc` and `WrapAction` methods for some different arities. Lots more to do - should these be made by hand or with code generation?
   - An alternative API could drop all of the new methods on `Instance` and instead expect the user to call these wrappers directly, e.g. `instance.GetFunction("swap")?.WrapFunc<int, int, (int, int)>()`
 - `IReturnTypeFactory` is pretty ugly - this uses reflection to construct various generic types for all the different arities of tuple without any boxing. Lots more to do - Should these be made by hand or with code generation?

## Code Generation

In a couple of cases a large list of very similar things needs generating (function wrappers and tuple factories). I don't know anything about the new code generation stuff in C#, so I could write these faster by hand. However it's probably more reliable and maintainable to generate them. What would you prefer?


peterhuene:
Hi @martindevans I've been heads down on other work this week, but I plan on taking a look at this early next week to see how viable this approach is so we can move forward on either of these approaches. I appreciate all your hard work on this!

peterhuene:
Hi @martindevans. I did a high-level review of the changes as I only wanted to comment on the approach rather than a full review.

I think there is great value in this design to elide the signature checks at the call site and I think we should move forward with this vs. just the value boxing.

> Function has an internal CheckTypeSignature method which simply checks if a given type signature matches this function. Could be useful to make public?

I don't see any reasons why it shouldn't be public; someone out there might find that useful.

> Function has various internal WrapFunc and WrapAction methods for some different arities. Lots more to do - should these be made by hand or with code generation? 

I've also never written a C# source generator, but looking over their implementation, it might be more effort for this particular use case than it's worth. However, if you feel that it would benefit from doing so, happy to accept it rather than hand-rolled boilerplate.

> An alternative API could drop all of the new methods on Instance and instead expect the user to call these wrappers directly, e.g. instance.GetFunction("swap")?.WrapFunc<int, int, (int, int)>()

In the Wasmtime Rust API, there are two basic ways to get a typed function: [`Func::wrap`](https://docs.rs/wasmtime/0.37.0/wasmtime/struct.Func.html#method.wrap) (used to create typed host functions for importing a function into an instance) and [`Func::typed`](https://docs.rs/wasmtime/0.37.0/wasmtime/struct.Func.html#method.typed) (used to "type" a function returned from `Instance::get_export`).  There is also a convenience method called [`Instance::get_typed_func`](https://docs.rs/wasmtime/0.37.0/wasmtime/struct.Func.html#method.wrap) that internally calls `Instance::get_export` and then `Instance::typed`.

I think perhaps we should follow the Rust API's lead and have the "wrap" methods on `Function` itself (and perhaps also used for creating the actual callback for the `FromCallback` overloads to elide the type checks for host functions too)? And then a convenience method on `Instance` called something like `GetWrappedFunction` that internally calls `Instance.GetFunction` followed by a call to wrap it?

What are your thoughts on that approach?

> IReturnTypeFactory is pretty ugly - this uses reflection to construct various generic types for all the different arities of tuple without any boxing. Lots more to do - Should these be made by hand or with code generation?

We have the same problem in the Rust implementation, but the Rust macro system is a little more flexible (albeit unwieldy) for generating things like this. I'd endorse the use of a source generator if you find it worthwhile. Otherwise, I'm happy to leave it hand-written.

That said, I don't think we'd ever want to go beyond being able to wrap up to 16 parameters and perhaps just a few results as multi-value returns are still not widely used in various toolsets; perhaps that can save us a bit of code in `IReturnTypeFactory` implementations.

I'm happy to give a full review when the PR leaves the draft state and I'll again take this opportunity to thank for your work on this high-value change.

martindevans:
Just a note on the progress of this: It was stalled due to the use of `byte[]` to represent the wasm `v128` type, I've just pushed a fix for this into the other PR (#114).

peterhuene:
I'm guessing the Windows packages for the latest upstream CI build failed to publish, so we can ignore those failures for now; likely will correct itself soon.

martindevans:
CI issues have resolved, and I think this is now ready for review 🥳

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// <summary>
+        /// Check if this function has the given type signature
+        /// </summary>
+        /// <param name="returnType">Return type (use a tuple for multiple return types)</param>
+        /// <param name="parameters">The parameters of the function</param>
+        /// <returns></returns>
+        public bool CheckTypeSignature(Type? returnType = null, params Type[] parameters)
+        {
+            // Check if the func returns no values if that's expected
+            if (Results.Count == 0 && returnType != null)
```
peterhuene, line 558: Nit: brackets around if body (here and throughout this method).

src/ValueBox.cs
```diff
@@ -169,5 +173,176 @@ public static ValueBox AsBox<T>(T? value)
         {
             return new ValueBox(value);
         }
+
+        internal static IValueBoxConverter<T> Converter<T>()
+        {
+            if (typeof(T) == typeof(int))
```
peterhuene, line 179: Nit: brackets around if statements in this method.

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
         /// </summary>
         public static Function Null => _null;
 
+        /// <summary>
+        /// Check if this function has the given type signature
+        /// </summary>
+        /// <param name="returnType">Return type (use a tuple for multiple return types)</param>
+        /// <param name="parameters">The parameters of the function</param>
+        /// <returns></returns>
```
peterhuene, line 554:
```suggestion
        /// <returns>Returns true if the type signature of the function is valid or false if not.</returns>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+            // All ok!
+            return true;
+        }
+
+        #region wrap
+        /// <summary>
+        /// Attempt to wrap this function as an Action. Wrapped action is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 626:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+                InvokeWithoutReturn(store, args);
+            };
+        }
+
+        /// <summary>
+        /// Attempt to wrap this function as an Action. Wrapped action is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 653:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+            };
+        }
+
+        /// <summary>
+        /// Attempt to wrap this function as an Action. Wrapped action is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 683:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        }
+
+        /// <summary>
+        /// Attempt to wrap this function as an Action. Wrapped action is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 716:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+
+        /// <summary>
+        /// Attempt to wrap this function as an Action. Wrapped action is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <typeparam name="TE">Fifth parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 752:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// <summary>
+        /// Attempt to wrap this function as an Action. Wrapped action is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <typeparam name="TE">Fifth parameter</typeparam>
+        /// <typeparam name="TF">Sixth parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 792:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// Attempt to wrap this function as an Action. Wrapped action is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <typeparam name="TE">Fifth parameter</typeparam>
+        /// <typeparam name="TF">Sixth parameter</typeparam>
+        /// <typeparam name="TG">Seventh parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 835:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// </summary>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <typeparam name="TE">Fifth parameter</typeparam>
+        /// <typeparam name="TF">Sixth parameter</typeparam>
+        /// <typeparam name="TG">Seventh parameter</typeparam>
+        /// <typeparam name="TH">Eighth parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 881:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <typeparam name="TE">Fifth parameter</typeparam>
+        /// <typeparam name="TF">Sixth parameter</typeparam>
+        /// <typeparam name="TG">Seventh parameter</typeparam>
+        /// <typeparam name="TH">Eighth parameter</typeparam>
+        /// <typeparam name="TI">Ninth parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 930:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+
+                InvokeWithoutReturn(store, args);
+            };
+        }
+
+        /// <summary>
+        /// Attempt to wrap this function as a Func. Wrapped func is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TR">Return type</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 974:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+                return InvokeWithReturn(store, stackalloc Value[0], factory);
+            };
+        }
+
+        /// <summary>
+        /// Attempt to wrap this function as a Func. Wrapped func is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TR">Return type</typeparam>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 998:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+            };
+        }
+
+        /// <summary>
+        /// Attempt to wrap this function as a Func. Wrapped func is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TR">Return type</typeparam>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">First parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 1028:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+
+        /// <summary>
+        /// Attempt to wrap this function as a Func. Wrapped func is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TR">Return type</typeparam>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 1099:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// <summary>
+        /// Attempt to wrap this function as a Func. Wrapped func is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TR">Return type</typeparam>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <typeparam name="TE">Fifth parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 1139:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// Attempt to wrap this function as a Func. Wrapped func is faster than a normal Invoke call.
+        /// </summary>
+        /// <typeparam name="TR">Return type</typeparam>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <typeparam name="TE">Fifth parameter</typeparam>
+        /// <typeparam name="TF">Sixth parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 1182:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// </summary>
+        /// <typeparam name="TR">Return type</typeparam>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <typeparam name="TE">Fifth parameter</typeparam>
+        /// <typeparam name="TF">Sixth parameter</typeparam>
+        /// <typeparam name="TG">Seventh parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 1228:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// <typeparam name="TR">Return type</typeparam>
+        /// <typeparam name="TA">First parameter</typeparam>
+        /// <typeparam name="TB">Second parameter</typeparam>
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <typeparam name="TE">Fifth parameter</typeparam>
+        /// <typeparam name="TF">Sixth parameter</typeparam>
+        /// <typeparam name="TG">Seventh parameter</typeparam>
+        /// <typeparam name="TH">Eighth parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 1277:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// <typeparam name="TC">Third parameter</typeparam>
+        /// <typeparam name="TD">Fourth parameter</typeparam>
+        /// <typeparam name="TE">Fifth parameter</typeparam>
+        /// <typeparam name="TF">Sixth parameter</typeparam>
+        /// <typeparam name="TG">Seventh parameter</typeparam>
+        /// <typeparam name="TH">Eighth parameter</typeparam>
+        /// <typeparam name="TI">Ninth parameter</typeparam>
+        /// <typeparam name="TJ">Tenth parameter</typeparam>
+        /// <typeparam name="TK">Eleventh parameter</typeparam>
+        /// <param name="store"></param>
```
peterhuene, line 1444:
```suggestion
        /// <param name="store">The store to use when calling the function.</param>
```

peterhuene (commented):
This looks fantastic.

Almost every comment I have is a formatting nit or missing doc comment.

The implementation itself looks really solid and I'm really happy with how this turned out!

src/Function.cs
```diff
@@ -662,10 +2010,58 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+            {
+                throw new ArgumentNullException(nameof(store));
+            }
+
+            var context = store.Context;
+
+            IntPtr error;
+            IntPtr trap;
+            fixed (Value* argsPtr = arguments)
+            fixed (Value* resultsPtr = resultsOut)
```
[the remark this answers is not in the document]
martindevans: I've done them as a combined brace for both, instead of a double level of indentation. Just say if you'd prefer both :)

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        /// <summary>
+        /// Invokes the wasmtime function and processes the results through a return type factory.
+        /// Assumes arguments are the correct type. Disposes the arguments.
+        /// </summary>
+        /// <typeparam name="TR"></typeparam>
+        /// <param name="store"></param>
+        /// <param name="arguments"></param>
+        /// <param name="factory"></param>
+        /// <returns></returns>
+        internal unsafe TR InvokeWithReturn<TR>(IStore store, ReadOnlySpan<Value> arguments, IReturnTypeFactory<TR> factory)
```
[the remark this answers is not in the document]
martindevans: I've made this private but left on the doc comment, since it includes details of some fairly dangerous assumptions the function makes.

src/Function.cs
```diff
@@ -546,6 +546,1365 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+        }
+
+        /// <summary>
+        /// Invokes the wasmtime function
+        /// Assumes arguments are the correct type. Disposes the arguments.
+        /// </summary>
+        /// <param name="store"></param>
+        /// <param name="arguments"></param>
+        /// <returns></returns>
+        internal unsafe void InvokeWithoutReturn(IStore store, ReadOnlySpan<Value> arguments)
```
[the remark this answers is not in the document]
martindevans: I've made this private but left on the doc comment, since it includes details of some fairly dangerous assumptions the function makes.

src/Function.cs
```diff
@@ -662,10 +2010,58 @@ public static Function FromCallback<TResult>(IStore store, Func<TResult> callbac
...
+            {
+                for (int i = 0; i < arguments.Length; ++i)
+                {
+                    args[i].Dispose();
+                }
+            }
+
+        }
+
+        /// <summary>
```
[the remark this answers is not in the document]
martindevans: Usually I wouldn't for private methods, but since these ones make some dangerous assumptions I think it's better to leave them in this case.

martindevans:
That should address all of the issues raised in review :)

martindevans:
Those failures looks like the CI is trying to download files that don't exist in the release again. No x86_86 build for Linux or Mac listed here: https://github.com/bytecodealliance/wasmtime/releases/tag/dev

peterhuene:
Yeah, looks like Wasmtime's CI is getting rate limited: https://github.com/bytecodealliance/wasmtime/actions/runs/2579496535.

I think we can safely ignore the CI failures.

src/Function.cs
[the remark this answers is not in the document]
peterhuene: Works for me, thanks!

peterhuene (approved):

peterhuene:
@martindevans I really appreciate all the effort you put into this!

I need to release 0.38.0 (I was sick last week) and 0.38.1 (released yesterday to fix a Cranelift bug), which I think can include these changes since they didn't touch any of the interop stuff (or have a dependency on an upstream wasmtime change).

martindevans:
Thanks for your support in giving these detailed reviews. I'm sure I'll be back to make more PRs soon ;)

Issue #117, opened 2022-06-10
wasmtime_config_interruptable_set missing


redthing1:
```
System.EntryPointNotFoundException: Unable to find an entry point named 'wasmtime_config_interruptable_set' in shared library 'wasmtime'.
   at Wasmtime.Config.Native.wasmtime_config_interruptable_set(Handle config, Boolean enable)
   at Wasmtime.Config.WithInterruptability(Boolean enable)

```

redthing1:
https://github.com/bytecodealliance/wasmtime/commit/c22033bf9397e9bec3704ff10ca897598e3dcd08

PR #118, opened 2022-06-10
add epoch interruption support

redthing1:
fixes: https://github.com/bytecodealliance/wasmtime-dotnet/issues/117

wasmtime removed and replaced old interruptability: https://github.com/bytecodealliance/wasmtime/commit/c22033bf9397e9bec3704ff10ca897598e3dcd08

also add a working unit test and features for this example: https://github.com/bytecodealliance/wasmtime/blob/main/examples/interrupt.c

### DIFF
--- a/src/Store.cs
+++ b/src/Store.cs
@@ -61,12 +61,16 @@ namespace Wasmtime
             }
         }
 
+        /// <summary>
+        /// Configures the relative deadline at which point WebAssembly code will trap.
+        /// </summary>
+        /// <param name="deadline"></param>
         public void SetEpochDeadline(ulong deadline)
         {
             Native.wasmtime_context_set_epoch_deadline(handle, deadline);
         }
 
-        public static class Native
+        private static class Native
         {
             [DllImport(Engine.LibraryName)]
             public static extern void wasmtime_context_gc(IntPtr handle);
@@ -85,7 +89,7 @@ namespace Wasmtime
             public static extern IntPtr wasmtime_context_set_wasi(IntPtr handle, IntPtr config);
             
             [DllImport(Engine.LibraryName)]
-            public static extern void wasmtime_context_set_epoch_deadline(IntPtr handle, ulong ticks_beyond_current);
+            public static extern void wasmtime_context_set_epoch_deadline(IntPtr handle, ulong ticksBeyondCurrent);
         }
 
         internal readonly IntPtr handle;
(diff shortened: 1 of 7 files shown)


src/Store.cs
```diff
@@ -61,7 +61,12 @@ internal void SetWasiConfiguration(WasiConfiguration config)
             }
         }
 
-        private static class Native
+        public void SetEpochDeadline(ulong deadline)
```
peterhuene, line 64: Also a doc comment here for this public method.

src/Store.cs
```diff
@@ -61,7 +61,12 @@ internal void SetWasiConfiguration(WasiConfiguration config)
...
+        {
+            Native.wasmtime_context_set_epoch_deadline(handle, deadline);
+        }
+
+        public static class Native
```
peterhuene, line 69: I think this inner class should probably remain `private` or at the very least `internal`.

src/Store.cs
```diff
@@ -232,6 +240,12 @@ public Store(Engine engine)
         /// </summary>
         /// <param name="config">The WASI configuration to use.</param>
         public void SetWasiConfiguration(WasiConfiguration config) => ((IStore)this).Context.SetWasiConfiguration(config);
+        
+        /// <summary>
+        /// Configures the relative deadline at which point WebAssembly code will trap.
+        /// </summary>
+        /// <param name="ticks_beyond_current"></param>
+        public void SetEpochDeadline(ulong ticks_beyond_current) => ((IStore)this).Context.SetEpochDeadline(ticks_beyond_current);
```
peterhuene, line 248: Naming convention nit: can we rename `ticks_beyond_current` to `ticksBeyondCurrent` for the [C# coding convention](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#camel-case)?

tests/EpochInterruptionTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        var run = instance.GetFunction(Store, "run");
+
+        var action = () => {
+            using (var timer = new Timer(state => Fixture.Engine.IncrementEpoch())) {
+                try {
+                    timer.Change(TimeSpan.FromMilliseconds(100), TimeSpan.FromMilliseconds(Timeout.Infinite));
+                    run.Invoke(Store);
+                }
+                catch (TrapException trap) {
+                    throw new TimeoutException("Invocation timed out", trap);
```
peterhuene, line 46: Rather than catching `TrapException` here and rethrowing a `TimeoutException`, can we just assert that the action throws `TrapException` and perhaps verify the exception message relates to hitting the deadline (I'd have to go look at Wasmtime to figure out what the trap message is in this case, though).

peterhuene (commented):
Hi @redthing1. Thanks very much for this PR! I have just a few comments to go through, minor things, otherwise this looks great 👍

redthing1:
Done @peterhuene

peterhuene (approved):

peterhuene:
@redthing1 thanks again for providing this PR!

PR #123, opened 2022-06-29
Add `GetAction` that returns `Action` without any parameters.

peterhuene:
This PR adds `Instance.GetAction` that takes no generic parameters to
return an `Action` for functions that take no parameters and return no results.

Adds `Function.WrapAction` that takes no generic parameters in kind.

Updates the README and examples to use the new wrapping functionality.


peterhuene:
@martindevans would you mind reviewing this? As I was updating the examples I realized we didn't add a `GetAction` that will work for functions that have no parameters or results, so i quickly added it and updated the examples.

martindevans:
This looks ok to me 👍 

I'd recommend adding a test invoking the new variant just in case - seems like an obvious place an edge case bug could be hiding!

PR #130, opened 2022-06-29
Type-Safe/Non-Boxing Global

martindevans:
Used the same mechanisms implemented for type-safe/non-boxing functions to wrap `Global` in the same way. I expected this to take longer to implement, but everything just fell into place perfectly 🎉

### DIFF
--- a/src/Global.cs
+++ b/src/Global.cs
@@ -194,16 +194,14 @@ namespace Wasmtime
         {
             private readonly Global _global;
             private readonly IStore _store;
-
-            private readonly IReturnTypeFactory<T> _factory;
+            
             private readonly IValueBoxConverter<T> _converter;
 
             internal Accessor(Global global, IStore store)
             {
                 _global = global ?? throw new ArgumentNullException(nameof(global));
                 _store = store ?? throw new ArgumentNullException(nameof(store));
-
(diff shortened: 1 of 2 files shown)


martindevans:
Some open questions with how this is implemented:

 - Name: Currently it's `Global.Accessor<T>` but `Global.Wrapper<T>` may be more consistent?
 - Mutability: Currently `Set(T)` does a runtime check and throws if the global is immutable. It could be written to instead have `WrapGetter<T>` and WrapSetter<T>` methods, with `WrapSetter` returning null if the global is immutable.

src/Global.cs
```diff
@@ -169,5 +184,72 @@ internal static class Native
         }
 
         private readonly ExternGlobal global;
+
+        /// <summary>
+        /// A typed accessor for a WebAssembly global value.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
```
peterhuene, line 191:
```suggestion
        /// <typeparam name="T">Type of the global being accessed.</typeparam>
```

martindevans:
I've filled in that doc comment :)

peterhuene:
Ignoring transient CI failure.

peterhuene:
@martindevans thanks again for all these great contributions!

PR #131, opened 2022-07-05
Module.Wat2Wasm

martindevans:
Added `Wat2Wasm` method, exposing the wasmtime `wat2wasm` function.

### DIFF
--- a/src/Module.cs
+++ b/src/Module.cs
@@ -320,7 +320,7 @@ namespace Wasmtime
         /// Convert WAT (Web Assembly Text) into WASM bytes
         /// </summary>
         /// <param name="wat">A string containing WAT (Web Assembly Text)</param>
-        /// <returns></returns>
+        /// <returns>Returns a byte array containing the WebAssembly module represented by the given text.</returns>
         /// <exception cref="ArgumentNullException">Thrown if text is null</exception>
         /// <exception cref="WasmtimeException">Thrown if text is not valid WAT</exception>
         public static byte[] ConvertText(string wat)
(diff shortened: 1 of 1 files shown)


src/Module.cs
```diff
@@ -316,6 +316,43 @@ public static Module DeserializeFile(Engine engine, string name, string path)
             return new Module(handle, name);
         }
 
+        /// <summary>
+        /// Convert WAT (Web Assembly Text) into WASM bytes
+        /// </summary>
+        /// <param name="text"></param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        public static byte[] Wat2Wasm(string text)
```
peterhuene, line 325: I don't think we have to limit ourselves to the underlying C API name here. Perhaps `ConvertText` would be a better name for C# (open to other names too)?

martindevans:
I've made those two suggested changes :)

src/Module.cs
```diff
@@ -316,6 +316,44 @@ public static Module DeserializeFile(Engine engine, string name, string path)
             return new Module(handle, name);
         }
 
+        /// <summary>
+        /// Convert WAT (Web Assembly Text) into WASM bytes
+        /// </summary>
+        /// <param name="wat">A string containing WAT (Web Assembly Text)</param>
+        /// <returns></returns>
```
peterhuene, line 323:
```suggestion
        /// <returns>Returns a byte array containing the WebAssembly module represented by the given text.</returns>
```

peterhuene (approved):

src/Module.cs
martindevans: Oops, I should have filled that one in myself. Sorry about that 🤦‍♂️

peterhuene:
@martindevans thanks for this! I'll merge when CI goes green.

PR #132, opened 2022-07-10
V128.AsSpan()

martindevans:
Added an `AsSpan` method to `V128` to allow for directly reading and writing the bytes of the `V128`. Without this `V128` is not very useful, since the value is inaccessible!


peterhuene (approved):
Thanks for this!

PR #134, opened 2022-07-15
Configurable string encodings (#133)

### LINKED ISSUE #133: Optional String Encoding of Memory
The memory layout of WASM strings compiled with AssemblyScript is in UTF16 format, while the memory.readstring functions are all hard coded to UTF8. https://github.com/AssemblyScript/assemblyscript/issues/1653

Providing an optional override for this format would be helpful, the temporary workaround is to encode strings within wasm, but ideally this could be avoided.

Joshhua5:
Added the ability to optionally override the encoding on string operations from Memory.

### DIFF
--- a/src/Memory.cs
+++ b/src/Memory.cs
@@ -64,6 +64,11 @@ namespace Wasmtime
         public uint Maximum { get; private set; }
 
         /// <summary>
+        /// The memory encoding for string operations.
+        /// </summary>
+        public Encoding MemoryEncoding { get; set; } = Encoding.UTF8;
+
+        /// <summary>
         /// Gets the current size of the memory, in WebAssembly page units.
         /// </summary>
         /// <returns>Returns the current size of the memory, in WebAssembly page units.</returns>
@@ -100,10 +105,10 @@ namespace Wasmtime
         /// <returns>Returns the span of the memory.</returns>
         /// <remarks>
         /// The span may become invalid if the memory grows.
-        /// 
+        ///
         /// This may happen if the memory is explicitly requested to grow or
         /// grows as a result of WebAssembly execution.
-        /// 
+        ///
         /// Therefore, the returned span should not be used after calling the grow method or
         /// after calling into WebAssembly code.
         /// </remarks>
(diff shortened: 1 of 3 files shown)


src/Memory.cs
```diff
@@ -62,6 +62,11 @@ public Memory(IStore store, uint minimum = 0, uint maximum = uint.MaxValue)
         /// </summary>
         public uint Maximum { get; private set; }
 
+        /// <summary>
+        /// The memory encoding for string operations. 
+        /// </summary>
+        public Encoding MemoryEncoding { get; set; } = Encoding.UTF8;
```
peterhuene, line 68: Would perhaps a `Encoding` parameter to `ReadString` and `WriteString` defaulted to `Encoding.UTF8` would be a little more ergonomic for anyone wishing to read different string encodings from the same linear memory (although not as likely as scenario)?

peterhuene: Related, we should update the doc comments for `ReadString` and `WriteString` so as to not be UTF-8 specific.

src/Memory.cs
```diff
@@ -131,7 +136,7 @@ public string ReadNullTerminatedString(IStore store, int address)
                 throw new InvalidOperationException("string is not null terminated");
             }
 
-            return Encoding.UTF8.GetString(slice.Slice(0, terminator));
+            return MemoryEncoding.GetString(slice.Slice(0, terminator));
```
peterhuene, line 139: I think we should keep `ReadNullTerminatedString` specific to UTF-8 given that we can't can't simply `IndexOf` for the null byte in other encodings.

peterhuene (commented):
Hi @Joshhua5. Thanks for putting this PR up!

Just a few review comments to go over.

peterhuene:
Hi @Joshhua5! I just wanted to touch base and see if you saw the feedback to your PR.

peterhuene:
Hi @Joshhua5. As I haven't heard from you in a while, I went ahead and amended your PR to include the requested changes.

As you did not create a local branch for your changes, I unfortunately had to push to your `main` branch to update the PR.

When this PR merges, please run `git fetch upstream && git reset --hard upstream/main` to hard reset your main branch to upstream. This assumes you've set up a remote named `upstream` for `git@github.com:bytecodealliance/wasmtime-dotnet.git`, though.

peterhuene (approved):

peterhuene:
@Joshhua5 thanks again for working on these changes!

Joshhua5:
@peterhuene Thank you for following through with this PR, it would be good for me to have notifications turn on.

PR #139, opened 2022-08-10
Pull aarch64 for macOS

willhausman:
@peterhuene I can setup an additional item to publish, but the output lib is identical for aarch64-macos and x86_64-macos, `libwasmtime.dylib`.  This results in one overriding the other in the test bin.

I tried changing the `<Content Include="..."><Link>%(WasmtimeDownload.WasmtimeLibraryFilename)</Link></Content>` to be `<Link>%(WasmtimeDownload.PackagePath)</Link>` so that the lib would be unique when copied to the output. That correctly gave me 4 libs, separated into their own runtime folders, but the tests could no longer pass. They all say the lib couldn't be found.

Have any thoughts?


peterhuene:
Hi @willhausman. I've put up #140 to address this issue. As I don't have an M1 mac, would you mind testing that PR to see if it works for you?

Thanks!

willhausman:
Closing in favor of #140.

PR #142, opened 2022-08-16
Generic Memory Methods

martindevans:
In my use of wasmtime I've often found myself doing something like this:

```csharp
var ptr = MemoryMarshal.Cast<byte, T>(_memory.GetSpan(Store!)[addr..]);
var v = ptr[0]; // Read
ptr[0] = new T(); // Write
```

to access a single value at an address.

I've added `Read<T>`, `Write<T>` and `GetSpan<T>` methods to `Memory` to make this pattern cleaner:

```csharp
var v = _memory.Read<T>(Store, addr);
_memory.Write<T>(Store, addr);
```

### DIFF
--- a/src/Memory.cs
+++ b/src/Memory.cs
@@ -108,7 +108,7 @@ namespace Wasmtime
         /// Gets the span of the memory viewed as a specific type, starting at a given address.
         /// </summary>
         /// <param name="store">The store that owns the memory.</param>
-        /// /// <param name="address">The zero-based address of the start of the span.</param>
+        /// <param name="address">The zero-based address of the start of the span.</param>
         /// <returns>Returns the span of the memory.</returns>
         /// <remarks>
         /// The span may become invalid if the memory grows.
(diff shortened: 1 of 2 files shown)


src/Memory.cs
```diff
@@ -104,6 +104,53 @@ public unsafe Span<byte> GetSpan(IStore store)
             return new Span<byte>(data, size);
         }
 
+        /// <summary>
+        /// Gets the span of the memory viewed as a specific type, starting at a given address.
+        /// </summary>
+        /// <param name="store">The store that owns the memory.</param>
+        /// /// <param name="address">The zero-based address of the start of the span.</param>
```
peterhuene, line 111:
```suggestion
        /// <param name="address">The zero-based address of the start of the span.</param>
```

peterhuene (approved):
Looks great, just the little fix is needed for the doc comment for `GetSpan<T>`. Thanks for adding this!

PR #143, opened 2022-08-19
Memory IStore Reference

martindevans:
Storing a reference to the `IStore` which was used to create the `Memory` object, this simplifies the API (all instance methods no longer need to be given an `IStore` reference) and removes a potential footgun (using the wrong `IStore`).

If you accept this PR then I can make the same change to `Function`, `Global` and `Table` as well.


martindevans:
CI failure is my fault, I was missing that test file for some reason.

peterhuene (approved):
This is so much better, thank you!

PR #145, opened 2022-08-24
Global Store Reference

martindevans:
Modified `Global` to keep a reference to the store which owns it. Much like #143 this simplifies the API and removes the potential footgun of trying to use a Global with the wrong Store


peterhuene (approved):
👍

PR #146, opened 2022-08-24
Function Store Reference

martindevans:
Modified Table to keep a reference to the store which owns it.

This required a change to the `ValueBox` and `ReturnTypeFactory` to take an `IStore` instead of a `StoreContext` (so that when a function is unboxed there's an `IStore` available to construct it).


peterhuene:
Looks like this needs to be rebased from the other two merges.

peterhuene (approved):
Looks great, just the XML doc comment to remove!

martindevans:
For some reason the CI hasn't started a new run after pushing that up :/

Issue #150, opened 2022-09-14
Overflow Exception In Trap Handling


martindevans:
I'm experimenting with a better way to handle traps and I've come across what looks like a bug in the current implementation.

I modified `Trap.wat` to cause a stackoverflow instead of using `unreachable`. I expected that to cause a `TrapException` of type `StackOverflow`, instead I'm seeing an `OverflowException`:

```csharp
Xunit.Sdk.XunitException
Expected a <Wasmtime.TrapException> to be thrown, but found <System.OverflowException>: "
"System.OverflowException with message "Arithmetic operation resulted in an overflow."
     at System.UIntPtr.op_Explicit(UIntPtr value)
     at Wasmtime.TrapFrame..ctor(IntPtr frame) in C:\Users\Martin\Documents\dotnet\wasmtime-dotnet\src\TrapException.cs:line 50
     at Wasmtime.TrapException.FromOwnedTrap(IntPtr trap) in C:\Users\Martin\Documents\dotnet\wasmtime-dotnet\src\TrapException.cs:line 179
     at Wasmtime.Function.Invoke(ReadOnlySpan`1 arguments, Span`1 resultsOut) in C:\Users\Martin\Documents\dotnet\wasmtime-dotnet\src\Function.cs:line 2218
     at Wasmtime.Function.InvokeWithoutReturn(ReadOnlySpan`1 arguments) in C:\Users\Martin\Documents\dotnet\wasmtime-dotnet\src\Function.cs:line 2063
```

This is coming from this code in the `TrapFrame` constructor:

```csharp
FunctionOffset = (int)Native.wasm_frame_func_offset(frame); // <-- Exception Here!
FunctionName = null;
ModuleOffset = (int)Native.wasm_frame_module_offset(frame);
ModuleName = null;
```

`Native.wasm_frame_func_offset` returns a `UIntPtr` (platform dependent, 64 bits on my PC) but it's being cast into an `int` (always 32 bits).

Should `FunctionOffset` be converted into an `nuint` instead? If so should the same change be made to `ModuleName`?

martindevans:
The overflow exception was fixed in #153, which included the change to use `nuint` instead (see [here](https://github.com/bytecodealliance/wasmtime-dotnet/blob/main/src/TrapException.cs#L71)).

peterhuene:
Thanks for verifying this!

PR #151, opened 2022-09-14
Trap Result

martindevans:
Added in 4 new `Result` types (`Result`, `ResultWithBacktrace`, `Result<T>`, `ResultWithBacktrace<T>`) which provide a better method for trap handling than exceptions. This entire new system is opt-in, the function wrappers accept these types instead of the actual return type and then, if a trap happens, it is returned in this result struct instead of as an exception.

 - `Result` represents a function with no return value, which does not capture the trap backtrace.
 - `ResultWithBacktrace` represents a function with no return value, which captures the trap backtrace.
 - `Result<T>` represents a function which returns `T`, which does not capture the trap backtrace.
 - `ResultWithBacktrace<T>` represents a function which returns `T`, which captures the trap backtrace.

Also some associated to the `TrapException`:
 - Added a `TrapCode Type` property, which was an obvious omission from traps
 - Modified `FunctionOffset` and `ModuleOffset` of the `TrapFrame` to use `nuint` instead of `int` (which could cause an arithmetic overflow) (see #150)

### DIFF
--- a/tests/TrapTests.cs
+++ b/tests/TrapTests.cs
@@ -46,6 +46,56 @@ namespace Wasmtime.Tests
                 .WithMessage("wasm trap: wasm `unreachable` instruction executed*");
         }
 
+        [Fact]
+        public void ItReturnsATrapCodeResult()
+        {
+
+            var instance = Linker.Instantiate(Store, Fixture.Module);
+            var run = instance.GetFunction<Result>("run_stack_overflow");
+            var result = run();
+
+            result.Type.Should().Be(ResultType.Trap);
+            result.Trap.Should().Be(TrapCode.StackOverflow);
+        }
+
+        [Fact]
+        public void ItReturnsATrapCodeAndBacktraceResult()
+        {
+
+            var instance = Linker.Instantiate(Store, Fixture.Module);
+            var run = instance.GetFunction<ResultWithBacktrace>("run_stack_overflow");
+            var result = run();
+
+            result.Type.Should().Be(ResultType.Trap);
+            result.Trap.Type.Should().Be(TrapCode.StackOverflow);
+            result.Trap.Frames.Count.Should().BeGreaterThan(0);
+        }
+
+        [Fact]
+        public void ItReturnsATrapCodeGenericResult()
+        {
+
+            var instance = Linker.Instantiate(Store, Fixture.Module);
+            var run = instance.GetFunction<Result<int>>("run_stack_overflow_with_result");
+            var result = run();
+
+            result.Type.Should().Be(ResultType.Trap);
+            result.Trap.Should().Be(TrapCode.StackOverflow);
+        }
+
+        [Fact]
+        public void ItReturnsATrapCodeAndBacktraceGenericResult()
+        {
+
+            var instance = Linker.Instantiate(Store, Fixture.Module);
+            var run = instance.GetFunction<ResultWithBacktrace<int>>("run_stack_overflow_with_result");
+            var result = run();
+
+            result.Type.Should().Be(ResultType.Trap);
+            result.Trap.Type.Should().Be(TrapCode.StackOverflow);
+            result.Trap.Frames.Count.Should().BeGreaterThan(0);
+        }
+
         public void Dispose()
         {
             Store.Dispose();
--- a/src/Function.cs
+++ b/src/Function.cs
@@ -554,6 +554,20 @@ namespace Wasmtime
         /// <returns>Returns true if the type signature of the function is valid or false if not.</returns>
         public bool CheckTypeSignature(Type? returnType = null, params Type[] parameters)
         {
+            // Check if the return type is Result<T>, Result, ResultWithBacktrace<T> or ResultWithBacktrace
+            if (returnType != null && returnType.IsResult())
+            {
+                // Try to get the type the result wraps (may be null if it's one of the non-generic result types)
+                var wrappedReturnType = returnType.IsGenericType ? returnType.GetGenericArguments()[0] : null;
+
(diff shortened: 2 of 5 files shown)


martindevans:
I'm not sure what the macos failure is caused by, it looks like something wrong with the CI process itself?

willhausman:
That error looks an awful lot like the crashes I would see when testing throwing exceptions in WasmerSharp and Wasmtime to compare. It happened locally on mac but not windows. Frustrating to discover. I only mention it to caution against assuming the CI itself is at fault.

martindevans:
Unfortunately I don't have access to a mac to test on. Any hints on what fixed the problem for you that I could try?

willhausman:
I can try pulling your fork and see what happens for me later tonight or tomorrow. My solution was to stop using WasmerSharp. Useful, right? Ha. It just feels too much of a coincidence, given you are working with the traps, and I was testing how the two runtimes handled exceptions.

willhausman:
I can confirm `dotnet test Wasmtime.sln` does exhibit this same behavior for me locally on both x64 and ARM Macs. Sorry I can't provide any more useful advice beyond my test results at the moment. I understand the traps at a conceptual level, but have not sifted through the implementation thoroughly enough to understand exactly what is happening.

martindevans:
Thanks for the testing, that at least confirms it isn't just a flakey CI problem.

martindevans:
If possible could you share the log from the test on your Mac? Hopefully that will contain more useful information than the CI log.

martindevans:
I have worked around the CI failure on MacOS by replacing the stackoverflow errors which integer div zero errors. I think the problem with MacOS+Stackoverflow is a bug in wasmtime-dotnet, independent of the changes in this PR.

I've pushed up #152 which includes a test that I expect to fail on MacOS to help identify the issue.

peterhuene:
Hi @martindevans! I'll review this today and dig into the stack overflow trap crashing the test host.

tests/TrapTests.cs
```diff
@@ -46,6 +46,56 @@ public void ItIncludesAStackTrace()
...
+        [Fact]
+        public void ItReturnsATrapCodeResult()
+        {
+
+            var instance = Linker.Instantiate(Store, Fixture.Module);
+            var run = instance.GetFunction<Result>("run_div_zero");
+            var result = run();
+
+            result.Type.Should().Be(ResultType.Trap);
+            result.Trap.Should().Be(TrapCode.IntegerDivisionByZero);
```
peterhuene, line 58: Should we be asserting here (and in `ItReturnsATrapCodeGenericResult`) that the frames are empty?

peterhuene (commented):
This looks great and I love the opt-in nature of it!

Just a few comments; almost all of them are minor nits.

src/Result.cs
```diff
@@ -0,0 +1,270 @@
...
+    {
+        /// <summary>
+        /// Excecution succeeded
+        /// </summary>
+        Ok = 0,
+
+        /// <summary>
+        /// Result contains a trap
+        /// </summary>
+        Trap = 2,
```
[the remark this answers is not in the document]
martindevans: Oops! Originally I had an `Error = 1` variant but decided that leaving errors as exceptions made more sense.

tests/TrapTests.cs
martindevans: Unless I'm misunderstanding what you're asking for I don't think that's possible in those two cases. They're both `Result` types (not `ResultWithBacktrace`) so the only thing available is the `TrapCode`, the backtrace is not available.

martindevans:
I've pushed up a commit that should address all feedback, except that last point. Thanks for the excellent review, as usual 👍

tests/TrapTests.cs
peterhuene: Oh right, the name `Trap` threw me off here. It should probably be, at the minimum, `TrapCode` so it doesn't seem like the whole trap exception is being stored.

Stepping back, though, should we not store a `TrapException` for the non-backtrace capturing `Result` types too, just without `Frames` populated?

I think it'd be useful to get the message in addition to the code, just without the backtrace.

martindevans: Good point on the naming, I'll change that now 👍

Regarding the `TrapException`/`message` my intention was to have the `Result` be as lightweight as possible, avoiding allocating a `TrapException` and the message string. Personally I would prefer to keep it that way.

Maybe the problem here is naming - `Result` (the simpler name) is the opt-in to a lighter weight version with no useful debugging information. Perhaps instead they should be named `Result` (with a backtrace) and `ResultWithoutAnyUsefulInformation` (pending a better name). That way it's probably more obvious that the version with the backtrace is the one you "should" use?

---

Having said all that, I'm wondering if another possible way to do this is instead of having some special "blessed types" which can interact with traps we should instead detect if the return type implements a special _interface_ and then do things through that. That way we can ship a single `Result` type with backtraces/messages etc, but there's a door to hook into the trap system (in a safe way of course) and get just what you need. If you don't mind holding off on this PR for a while I'll experiment with that approach to see if it's feasible.

peterhuene: Sounds good, I'll hold off on the PR to see what you come up with.

I'm also good with landing this PR if that's what you decide is best.

peterhuene (approved):

martindevans:
I've pushed up #153 which further develops this idea with a system for custom result types. Although the details of how it works are more complex I think it offers a better user experience for the 99% of users who can just use `ActionResult`/`FunctionResult<T>` and don't need to think about if they want a backtrace etc. It also offers a bettter experience for the remaining 1% of users with weird requirements, since they can implement fully custom result types instead of having to pick between whatever options we happen to offer.

PR #152, opened 2022-09-15
Trap Stackoverflow Test

martindevans:
My other PR (#151) was failing CI on MacOS only, I think this is due to a bug in wasmtime dotnet, although I'm not sure what and unfortunately I cannot debug it further since I don't own a Mac.

This PR includes the minimal set of changes to (hopefully) reproduce the problem.

Note that the `nuint` change is necessary to prevent an `OverflowException` (as mentioned in #150).

### DIFF
--- a/src/TrapException.cs
+++ b/src/TrapException.cs
@@ -47,9 +47,9 @@ namespace Wasmtime
     {
         unsafe internal TrapFrame(IntPtr frame)
         {
-            FunctionOffset = (int)Native.wasm_frame_func_offset(frame);
+            FunctionOffset = (nuint)Native.wasm_frame_func_offset(frame);
             FunctionName = null;
-            ModuleOffset = (int)Native.wasm_frame_module_offset(frame);
+            ModuleOffset = (nuint)Native.wasm_frame_module_offset(frame);
             ModuleName = null;
 
             var bytes = Native.wasmtime_frame_func_name(frame);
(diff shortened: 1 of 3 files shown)


martindevans:
As expected this PR has failed CI. As mentioned above I can't debug it any further, I hope this helps someone with access to a Mac debug the issue.

TomGillen:
I have been testing this, and found that `Native.wasmtime_func_call` crashes the test runner if the executed wasm overflows the stack. I have seen this happen on master, so this is a bug unrelated to this PR.

martindevans:
I've just put together a [test of this](https://github.com/martindevans/wasmtime-macos-test/blob/main/src/lib.rs#L51) in Rust which is fine when run on macos (using GH actions). So it seems to _not_ be a bug in the underlying library.

martindevans:
I've been doing even more testing with GH actions.

As you can see from this [execution log](https://github.com/martindevans/wasmtime-macos-test/actions/runs/3070035675/jobs/4959344970) [this program](https://github.com/martindevans/wasmtime-macos-test/blob/dotnet/ConsoleApp/ConsoleApp/Program.cs) works when you run it with:

> dotnet run --project ConsoleApp/ConsoleApp/ConsoleApp.csproj

However, the same [code copied into a test](https://github.com/martindevans/wasmtime-macos-test/blob/dotnet/wasmtime-dotnet/tests/TrapTests.cs#L66) and run with:

> dotnet test wasmtime-dotnet/tests/Wasmtime.Tests.csproj

Crashes!

I don't really have any ideas how to debug this further...

peterhuene:
I'll dive into this soon and figure out why it's failing in the test host; perhaps it is installing a vectored exception handler that is interfering with Wasmtime's.

peterhuene:
I'll still be looking into this issue, but I'm not holding up the 1.0.0 release for it as I believe it's a test-host only problem.

martindevans:
I'm going to close this, I don't think it's been relevant for a long time.

PR #153, opened 2022-09-21
Trap result builders

martindevans:
Created a new system to allow users of wasmtime-dotnet to create fully custom result types. This means that the library can ship a basic result type that fits 99% of usecases (just containing a `TrapException` object) and advanced users can build whatever they need for weirder usecases (e.g. making this work in Unity Jobs+Burst, which was my original goal for all this).

A custom result type implements one of two interfaces:
 - `IFunctionResult<,,>`, this represents a return value from a function (approximately equivalent to `Result<Whatever, Trap>` in Rust)
 - `IActionResult<,>`, this represents a return value from an action (approximately like returning `Option<Trap>` instead of `Result<(), Trap>` in Rust)

Unfortunately since `static abstract` methods aren't available yet (they're available soon, but would require a dependency on .NET7 afaik) the construction of these custom types is a little more complex than I'd like. The result interfaces contain a parameter refencing a "builder" interface which acts as a factory for the result type. You can see this in action in the `TrapTests` where I have implemented a very simple custom result:

```csharp
public struct MyCustomResult<T>
    : IFunctionResult<MyCustomResult<T>, T, MyCustomResult<T>.MyCustomResultBuilder>
{
    public int TrapStackDepth;

    public struct MyCustomResultBuilder
        : IFunctionResultBuilder<MyCustomResult<T>, T>
    {
        public MyCustomResult<T> Create(T value)
        {
            return new MyCustomResult<T> { TrapStackDepth = -1 };
        }

        public MyCustomResult<T> Create(TrapAccessor accessor)
        {
            return new MyCustomResult<T> { TrapStackDepth = accessor.GetFrames().Count };
        }
    }
}
```

### DIFF
--- a/src/Result.cs
+++ b/src/Result.cs
@@ -1,0 +1,270 @@
+﻿using System;
+
+namespace Wasmtime
+{
+    /// <summary>
+    /// Indicates what type of result this is
+    /// </summary>
+    public enum ResultType
+    {
+        /// <summary>
+        /// Excecution succeeded
+        /// </summary>
+        Ok = 0,
+
+        /// <summary>
+        /// Result contains a trap
+        /// </summary>
+        Trap = 2,
+    }
+
+    /// <summary>
+    /// A result from a function call which may represent a Value or a Trap. If a trap happens the full backtrace is captured.
+    /// </summary>
+    public readonly struct ResultWithBacktrace
+    {
+        /// <summary>
+        /// Indicates what type of result this contains
+        /// </summary>
+        public ResultType Type { get; }
+
+        private readonly TrapException? _trap;
+
+        internal ResultWithBacktrace(IntPtr trap)
+        {
+            Type = ResultType.Trap;
+            _trap = TrapException.FromOwnedTrap(trap);
+        }
+
+        /// <summary>
+        /// Get the trap associated with this result
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown if this Type != Types.Trap</exception>
+        public TrapException Trap
+        {
+            get
+            {
+                if (Type != ResultType.Trap)
+                    throw new InvalidOperationException($"Cannot get 'Trap' from '{Type}' type result");
+                return _trap!;
+            }
+        }
+    }
+
+    /// <summary>
+    /// A result from a function call which may represent a Value or a Trap
+    /// </summary>
+    public readonly struct Result
+    {
+        /// <summary>
+        /// Indicates what type of result this contains
+        /// </summary>
+        public ResultType Type { get; }
+
+        private readonly TrapCode _trap;
+
+        internal Result(IntPtr trap)
+        {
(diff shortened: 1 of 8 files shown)


martindevans:
> do you require this PR to be included in that release?

I plan to use this in our Unity wasm asset, but we're not on a tight timetable with that so if it slips to the next version I won't be too upset ;)

peterhuene:
Apologies for the delay, I'm reviewing this now.

src/TrapException.cs
```diff
@@ -128,65 +273,37 @@ public class TrapException : WasmtimeException
...
-                if (indexOfNull != -1)
-                {
-                    byteSpan = byteSpan.Slice(0, indexOfNull);
-                }
+                ExitCode = accessor.ExitStatus
+            };
 
-                var message = Encoding.UTF8.GetString(byteSpan);
-                bytes.Dispose();
+            if (delete)
```
peterhuene, line 299: Nit: could we put braces around this statement?

Additionally, I wonder if we should wrap the above in a `try`..`finally` to dispose the trap accessor (when `delete` is true) just so we don't accidentally leak the trap if something goes wrong.

src/Result.cs
```diff
@@ -0,0 +1,294 @@
...
+        /// Create an new instance indicating a successful call.
+        /// </summary>
+        /// <param name="value">The return value of the function call</param>
+        /// <returns>A new TResult instance representing a successful function call</returns>
+        TResult Create(TOk value);
+
+        /// <summary>
+        /// Create a new instance indicating a trap result.
+        /// </summary>
+        /// <param name="accessor"></param>
```
peterhuene, line 240: Nit: missing parameter doc comment.

src/TrapException.cs
```diff
@@ -101,6 +101,151 @@ private static class Native
...
+
+                    return totalBytes;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Get the trap frames
+        /// </summary>
+        /// <returns></returns>
```
peterhuene, line 220: Nit: missing return doc comment.

src/TrapException.cs
```diff
@@ -101,6 +101,151 @@ private static class Native
...
+                }
+
+                return trapFrames;
+            }
+        }
+
+        /// <summary>
+        /// Get a TrapException which contains all information about this trap
+        /// </summary>
+        /// <returns></returns>
```
peterhuene, line 242: Nit: missing parameter doc comment.

peterhuene (commented):
Thanks again for implementing this. I thinks this is a great improvement.

Just a few minor nits and a comment about protecting a `Dispose` call site with a try..finally.

Happy to merge this and get it into the 1.0 release, which I'll release this afternoon (PDT).

martindevans:
Thanks for getting this review done before the 1.0 deadline 👍

PR #161, opened 2022-10-10
Fix issues in `Function.FromCallback()` and `Linker.DefineFunction()`

### LINKED ISSUE #158: `Function.InvokeCallback()` should handle `ITuple` return value only when `Function.IsTuple()` was `true`

### LINKED ISSUE #159: `Function.InvokeCallback()` calls `callback.Method.Invoke()`, which can have different parameter count than the `Delegate`

kpreisser:
Hi, this is a PR to fix the following issues when using `Function.FromCallback()` and `Linker.DefineFunction()`:

- Instead of invoking `callback.Method` (that may have a different parameter count than the callback), invoke the callback's `Invoke` method.
- Handle an `ITuple` return value only when `GetFunctionType()` determined that the function returns a tuple, to correctly return a boxed `ValueTuple` as `externref`.
- Support nested levels of `ValueTuple`s, so that it's possible e.g. to return a `ValueTulple<..., ValueTuple<..., ValueTuple<...>>>` (15 values).

Fixes #158
Fixes #159

Thank you!

### DIFF
--- a/src/Function.cs
+++ b/src/Function.cs
@@ -2239,7 +2239,8 @@ namespace Wasmtime
             }
             this.store = store;
 
-            using var funcType = GetFunctionType(callback.GetType(), hasReturn, this.parameters, this.results, out var hasCaller);
+            using var funcType = GetFunctionType(callback.GetType(), hasReturn, this.parameters, this.results, out var hasCaller, out var returnsTuple);
+            var callbackInvokeMethod = callback.GetType().GetMethod(nameof(Action.Invoke))!;
 
             unsafe
             {
@@ -2249,7 +2250,7 @@ namespace Wasmtime
                     func = (env, callerPtr, args, nargs, results, nresults) =>
                     {
                         using var caller = new Caller(callerPtr);
-                        return InvokeCallback(callback, caller, true, args, (int)nargs, results, (int)nresults, Results);
+                        return InvokeCallback(callback, callbackInvokeMethod, caller, true, args, (int)nargs, results, (int)nresults, Results, returnsTuple);
                     };
                 }
                 else
@@ -2257,7 +2258,7 @@ namespace Wasmtime
                     func = (env, callerPtr, args, nargs, results, nresults) =>
                     {
                         using var caller = new Caller(callerPtr);
-                        return InvokeCallback(callback, caller, false, args, (int)nargs, results, (int)nresults, Results);
+                        return InvokeCallback(callback, callbackInvokeMethod, caller, false, args, (int)nargs, results, (int)nresults, Results, returnsTuple);
                     };
                 }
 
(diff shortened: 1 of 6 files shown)


peterhuene (approved):
Excellent, thanks for these fixes!

PR #166, opened 2022-10-15
Allow to access a `Memory` with more than 32767 pages

### LINKED ISSUE #165: `Memory` with more than 32767 pages cannot be accessed
Hi, consider the following C# program (.NET 6.0, using `Wasmtime 1.0.0`):

```cs

    engine,
    "hello",
    @"
(module 
    (memory $mem1 65536)
    (export ""memory"" (memory $mem1))
)");

using (var linker = new Linker(engine)) {

    var memory = instance.GetMemory("memory")!;

    try {
        memory.WriteInt16(10, 123);
        Console.WriteLine("Success");
    }
    catch (Exception ex) {
        Console.WriteLine(ex.ToString());
    }

    try {
        var memorySpan = memory.GetSpan();
        Console.WriteLine("Success");
    }
    catch (Exception ex) {
        Console.WriteLine(ex.ToString());
    }
}
```

**Actual behavior:** When running the program, an `OverflowException` occurs when the memory is accessed. E.g. with `Memory.WriteInt16()`:
```
System.OverflowException: Arithmetic operation resulted in an overflow.
   at System.UIntPtr.ToUInt32()
   at Wasmtime.Memory.GetSpan() in C:\Users\VPC\Desktop\WebAssembly-Test\wasmtime-dotnet\src\Memory.cs:line 92
   at Wasmtime.Memory.WriteInt16(Int32 address, Int16 value) in C:\Users\VPC\Desktop\WebAssembly-Test\wasmtime-dotnet\src\Memory.cs:line 228
   at Program.<<Main>$>g__TestBigMemoryCustomWasm|0_3() in C:\Users\VPC\Desktop\WebAssembly-Test\TestWasmtime\Program.cs:line 326
```

Or, when changing the memory definition to `(memory $mem1 32768)`, a similar `OverflowException` occurs:
```
System.OverflowException: Value was either too large or too small for an Int32.
   at System.Convert.ThrowInt32OverflowException()
   at System.Convert.ToInt32(UInt32 value)
   at Wasmtime.Memory.GetSpan() in C:\Users\VPC\Desktop\WebAssembly-Test\wasmtime-dotnet\src\Memory.cs:line 92
   at Wasmtime.Memory.WriteInt16(Int32 address, Int16 value) in C:\Users\VPC\Desktop\WebAssembly-Test\wasmtime-dotnet\src\Memory.cs:line 228
   at Program.<<Main>$>g__TestBigMemoryCustomWasm|0_3() in C:\Users\VPC\Desktop\WebAssembly-Test\TestWasmtime\Program.cs:line 326
```

The `Memory` can currently only be accessed by calling `GetSpan()` that returns a `Span<byte>` over the whole memory area.
Unfortunately, a `Span` uses an `Int32` length, so it can only have a max length of 2^31-1. Using a memory with 32768 or more pages would exceed that value for a `Span<byte>` (using a memory with 65536 pages would even exceed the storage of a 32-bit unsigned integer for the length, see also https://github.com/bytecodealliance/wasmtime/pull/3134).

I'm not sure of the best approach to solve this, but I could imagine that we may need new APIs using a `long` instead of `int` for addresses (like `WriteInt16(long address, short value)` etc.), and maybe a `Span<byte> GetSpan(long address, int length)` that allows to retrieve a `Span` for the given address and length (when the length doesn't exceed 2^31-1).
Additionally, to allow the caller to access the whole memory at once, we might need to allow retrieving a pointer (`IntPtr`) to the memory start.

(With the [`Memory64` proposal](https://github.com/WebAssembly/memory64/blob/main/proposals/memory64/Overview.md), a memory can even be larger than 4 GiB. As far is I understand it, it would theoretically support a memory length of 2^64 bytes, which might need to use an `ulong` as address to access more than 2^63 bytes, but `ulong` in .NET isn't CLS-compliant, and such a memory size is as of today only a theoretical situation, so I think using a `long` should be fine.)

What do you think?

Thanks!

kpreisser:
Hi, this PR adds new `Int64`-based APIs to `Memory` to allow to access memories with more than 32767 pages.

Fixes #165 

I added a new method `GetSpan<byte>(long address, int length)` as well as `GetSpan<T>(long address, int length)` to retrieve a `Span` starting at the given `Int64` address. Additionally, I added a `GetLength()` and `GetPointer()` methods for getting a raw pointer to the memory, and the memory length in bytes (in case you want to access the whole memory area at once, even if it exceeds a size of 2 GiB).

Additionally, for the `ReadXyz()`/`WriteXyz()` methods, I added new overloads taking an `Int64` address, and fixed the existing (`Int32`-based) methods to not throw an `OverflowException`.

I also added an `ObsoleteAttribute` to the `GetSpan()` method to indicate that it will throw an `OverflowException` when the memory has more than 32767 pages, with a recommendation to use the `GetSpan(long address, int length)` method instead.

Alternatively, we could also drop the old (`Int32`-based) APIs as breaking change.

What do you think?

Thanks!

### DIFF
--- a/src/Memory.cs
+++ b/src/Memory.cs
@@ -221,42 +221,10 @@ namespace Wasmtime
         /// Note that WebAssembly always uses little endian as byte order. On platforms 
         /// that use big endian, you will need to convert numeric values accordingly.
         /// </remarks>
-        public T Read<T>(int address)
-            where T : unmanaged
-        {
-            return GetSpan<T>(address, 1)[0];
-        }
-
-        /// <summary>
-        /// Read a struct from memory.
-        /// </summary>
-        /// <typeparam name="T">Type of the struct to read. Ensure layout in C# is identical to layout in WASM.</typeparam>
-        /// <param name="address">The zero-based address to read from.</param>
-        /// <returns>Returns the struct read from memory.</returns>
-        /// <remarks>
-        /// Note that WebAssembly always uses little endian as byte order. On platforms 
-        /// that use big endian, you will need to convert numeric values accordingly.
-        /// </remarks>
         public T Read<T>(long address)
             where T : unmanaged
         {
             return GetSpan<T>(address, 1)[0];
-        }
-
-        /// <summary>
-        /// Write a struct to memory.
-        /// </summary>
-        /// <typeparam name="T"></typeparam>
-        /// <param name="address">The zero-based address to read from.</param>
-        /// <param name="value">The struct to write.</param>
-        /// <remarks>
-        /// Note that WebAssembly always uses little endian as byte order. On platforms 
-        /// that use big endian, you will need to convert numeric values accordingly.
-        /// </remarks>
-        public void Write<T>(int address, T value)
-            where T : unmanaged
-        {
-            GetSpan<T>(address, 1)[0] = value;
         }
 
         /// <summary>
@@ -282,23 +250,6 @@ namespace Wasmtime
         /// <param name="length">The length of bytes to read.</param>
         /// <param name="encoding">The encoding to use when reading the string; if null, UTF-8 encoding is used.</param>
         /// <returns>Returns the string read from memory.</returns>
-        public string ReadString(int address, int length, Encoding? encoding = null)
-        {
-            if (encoding is null)
-            {
-                encoding = Encoding.UTF8;
-            }
-
-            return encoding.GetString(GetSpan(address, length));
-        }
-
-        /// <summary>
-        /// Reads a string from memory with the specified encoding.
-        /// </summary>
-        /// <param name="address">The zero-based address to read from.</param>
-        /// <param name="length">The length of bytes to read.</param>
-        /// <param name="encoding">The encoding to use when reading the string; if null, UTF-8 encoding is used.</param>
-        /// <returns>Returns the string read from memory.</returns>
         public string ReadString(long address, int length, Encoding? encoding = null)
         {
             if (encoding is null)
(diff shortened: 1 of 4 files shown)


src/Memory.cs
```diff
@@ -72,10 +72,40 @@ public uint GetSize()
...
+        /// <remarks>
+        /// The pointer may become invalid if the memory grows.
+        ///
+        /// This may happen if the memory is explicitly requested to grow or
+        /// grows as a result of WebAssembly execution.
+        ///
+        /// Therefore, the returned pointer should not be used after calling the grow method or
+        /// after calling into WebAssembly code.
+        /// </remarks>
+        /// <returns></returns>
```
peterhuene, line 97: Nit: missing doc comment.

src/Memory.cs
```diff
@@ -106,11 +157,58 @@ public unsafe Span<byte> GetSpan()
...
+            }
+
+            var context = store.Context;
+            var data = Native.wasmtime_memory_data(context.handle, this.memory);
+            var memoryLength = this.GetLength();
+
+            // Note: A Span<T> can span more than 2 GiB bytes if sizeof(T) > 1.
+            long byteLength = (long)length * sizeof(T);
+
+            if (address > memoryLength - byteLength)
```
peterhuene, line 208: Formatting nit: could you put braces around the if body?

src/Memory.cs
```diff
@@ -119,10 +217,30 @@ public unsafe Span<T> GetSpan<T>(int address)
...
+        /// <param name="address">The zero-based address to read from.</param>
+        /// <returns>Returns the struct read from memory.</returns>
+        /// <remarks>
+        /// Note that WebAssembly always uses little endian as byte order. On platforms 
+        /// that use big endian, you will need to convert numeric values accordingly.
+        /// </remarks>
+        public T Read<T>(long address)
+            where T : unmanaged
+        {
+            return GetSpan<T>(address, 1)[0];
```
peterhuene, line 243: Could we just remove these many new `long` overloads and change the `int`-based ones to `long` given that `int` implicitly converts to `long`?

peterhuene (commented):
Thanks for this fix too!

Just a few minor nits and a more general comment regarding all the new `long` overloads for the memory accessor methods.

src/Memory.cs
kpreisser: Yes, we could remove the `int`-based overloads as they are already replaced with `long`-based ones. However, while that would be a source-compatible change (since `int` implicitely converts to `long`), it would be a binary-incompatible change (it requires a recompilation of the assembly that calls the methods).

If `wasmtime-dotnet` uses a similar release cycle as `wasmtime` which always increments the major version, I think such a change should be OK.
Thanks!

src/Memory.cs
kpreisser: Fixed.

src/Memory.cs
kpreisser: The `int`-based overloads are now replaced by the `long` based overloads.

peterhuene (approved):

src/Memory.cs
peterhuene: Agree, I am more concerned with source compatibility given the way Wasmtime will be rev'ing version numbers. Thanks for making these changes.

peterhuene:
@kpreisser thanks for addressing this!

PR #169, opened 2022-10-22
Add support for 64-bit Memory

kpreisser:
Hi, this PR adds support for 64-bit memories, by adjusting the APIs to match the C API introduced with bytecodealliance/wasmtime#3153 and bytecodealliance/wasmtime#3182.

It contains the following changes:
- Add `Config.WithMemory64()` for enabling the Memory64 proposal.
- Change properties/parameters in `Memory`, `MemoryExport` and `MemoryImport` that work with pages (like `Minimum`, `Maximum`, `Grow()`) to use a `long` instead of `uint`. (I chose `long` instead of `ulong` because memory pages cannot exceed `2**48`, and because `ulong` isn't CLS-compliant.)
  (This will be a breaking change.)
  - Added an `Is64Bit` property that specifies whether the memory is a 64-bit memory.
- Add tests to verify the new APIs added with #166 are able to access a memory beyond the 4 GiB area, and that the new APIs correctly retrieve the memory type definitions.

Additionally, this fixes previously incorrect definitions of native functions `wasmtime_memory_size` and `wasmtime_memory_grow`, where the parameter incorrectly used `uint` instead of `ulong` after the changes introduced in the Wasmtime C API.
Also, I noticed that for native Config functions taking a C `bool` parameter, the `[MarshalAs]` attribute was missing, which meant these parameter were marshaled by .NET as 4-byte `BOOL` instead of an 1-byte value.

What do you think?

Thanks!

### DIFF
--- a/src/Config.cs
+++ b/src/Config.cs
@@ -373,46 +373,46 @@ namespace Wasmtime
             public static extern void wasm_config_delete(IntPtr config);
 
             [DllImport(Engine.LibraryName)]
-            public static extern void wasmtime_config_debug_info_set(Handle config, bool enable);
+            public static extern void wasmtime_config_debug_info_set(Handle config, [MarshalAs(UnmanagedType.I1)] bool enable);
 
             [DllImport(Engine.LibraryName)]
-            public static extern void wasmtime_config_epoch_interruption_set(Handle config, bool enable);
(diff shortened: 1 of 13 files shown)


peterhuene (approved):
These changes look great! Thanks for taking the opportunity to implement this feature along with the fixes.

PR #170, opened 2022-10-29
Fix the `bool` return type declaration for native Trap functions

### LINKED ISSUE #171: `TrapException.Type` incorrectly returns `TrapCode.StackOverflow` instead of `TrapCode.Undefined` for host traps
Hi, consider the following program using `Wasmtime 1.0.0` on Windows 10 x64:
```cs
    engine,
    "hello",
    @"
(module
    (import """" ""host_trap"" (func $host_trap))
    (export ""host_trap"" (func $host_trap))
)");

linker.Define("", "host_trap", Function.FromCallback(store, () => throw new Exception()));

var hostTrap = instance.GetFunction<ActionResult>("host_trap")!;
var result = hostTrap();

Console.WriteLine("Trap Type: " + result.Trap.Type);
```

**Expected behavior:**
```
Trap Type: Undefined
```

**Actual behavior:**
```
Trap Type: StackOverflow
```

This occurs because the return type for native function `wasmtime_trap_code` (as well as `wasmtime_trap_exit_status`) is incorrectly declared as `bool` without `[return: MarshalAs(UnmanagedType.I1)]`, which means .NET marshals the value as 4-byte `BOOL` instead of an 1-byte value:
https://github.com/bytecodealliance/wasmtime-dotnet/blob/acc50d5e5899b3ec702e8d0b418ac4e7ff925c98/src/TrapException.cs#L335-L339

The incorrect definitions appear to have been introduced with #95.

Thanks!

kpreisser:
The return type for native functions `wasmtime_trap_exit_status` and `wasmtime_trap_code` was incorrectly declared as `bool` without `[return: MarshalAs(UnmanagedType.I1)]`, which meant .NET marshaled the value as 4-byte `BOOL` instead of an 1-byte value.

For example, this caused `TrapException.Type` incorrectly being set to `TrapCode.StackOverflow` instead of `TrapCode.Undefined`, because the return value of `wasmtime_trap_code` was not correctly marshaled.

Fixes #171

Thanks!

### DIFF
--- a/src/TrapException.cs
+++ b/src/TrapException.cs
@@ -333,9 +333,11 @@ namespace Wasmtime
             public static extern void wasm_frame_vec_delete(in FrameArray vec);
 
             [DllImport(Engine.LibraryName)]
+            [return: MarshalAs(UnmanagedType.I1)]
             internal static extern bool wasmtime_trap_exit_status(IntPtr trap, out int exitStatus);
 
             [DllImport(Engine.LibraryName)]
+            [return: MarshalAs(UnmanagedType.I1)]
             internal static extern bool wasmtime_trap_code(IntPtr trap, out TrapCode exitCode);
         }
     }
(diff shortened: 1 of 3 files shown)


peterhuene (approved):
Thanks!

PR #172, opened 2022-10-29
Pass original .NET callback exception as `TrapException`'s inner exception 

### LINKED ISSUE #63: Improve debuggability by adding original host function exceptions to TrapException
Right now if a host function throws a .NET exception, the exception is caught and translated to a trap representation.

The trap representation only keeps the exception message.  When the trap bubbles back up to .NET, the trap is thrown as a `TrapException`.

Since only the message is kept, debugging the source of the host function exception can be difficult, especially from a log that prints only the `TrapException`.

When a trap representation is created, the original caught exception should be stored somewhere that can be utilized from where `TrapException` is thrown and passed in as the inner exception.

kpreisser:
Hi, this PR implements the functionality to pass a .NET exception thrown at a wasm-to-host transition (.NET callback) (which means the exception is converted to a trap) as inner exception of the created `TrapException` when the trap bubbles up to the next host-to-wasm transition, by using a `ThreadStatic` (thread-local) field to temporarily store the exception until the stack is unwound.

That way, the original .NET exception that caused a trap can be retrieved from `TrapException.InnerException`, e.g. for debugging purposes.

If the exception thrown at the .NET callback is already a `TrapException`, we will not store that exception itself, but its `InnerException`, so that when the trap bubbles through multiple levels of wasm-to-host transitions, the `InnerException` at the root host-to-wasm transition will still be the original .NET exception.

Fixes #63 

Note: The mechanism requires that on every location where a trap can occur due to calling wasm code, `TrapException.FromOwnedTrap()` is called when a trap actually occured, so that the thread-local field is cleared afterwards.

TODO:
- If #163 is merged before this PR, the exception-handling code in the generated code for `Function.FromCallback` and `Linker.DefineCallback` needs to be adjusted to call `Function.HandleCallbackException()`.

What do you think?

Thanks!

### DIFF
--- a/src/Function.cs
+++ b/src/Function.cs
@@ -2450,12 +2450,28 @@ namespace Wasmtime
             }
             catch (Exception ex)
             {
-                var bytes = Encoding.UTF8.GetBytes(ex.Message);
+                return HandleCallbackException(ex);
+            }
+        }
 
-                fixed (byte* ptr = bytes)
-                {
-                    return Native.wasmtime_trap_new(ptr, (UIntPtr)bytes.Length);
-                }
+        internal static unsafe IntPtr HandleCallbackException(Exception ex)
+        {
+            // Store the exception as trap cause, so that we can use it as the TrapException's
+            // InnerException when the trap bubbles up to the next host-to-wasm transition.
+            // If the exception is already a TrapException, we use that one's InnerException,
+            // even if it's null.
+            // Note: This code currently assumes that on every host-to-wasm transition where a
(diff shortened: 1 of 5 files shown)


kpreisser:
The tests are currently failing due to #171.

peterhuene:
We should be able to rebase this PR now.

peterhuene (approved):
Thanks for implementing this feature! This should definitely help users in debugging host functions that throw an exception.

PR #175, opened 2022-11-03
Update implementation and tests for the trap redesign

### LINKED ISSUE #173: CI is red
There were [breaking changes upstream](https://github.com/bytecodealliance/wasmtime/pull/5149) that impacted the C API, which are breaking the .NET CI tests.

kpreisser:
Hi, this updates the implementation and tests for the trap redesign (bytecodealliance/wasmtime#5149, https://github.com/bytecodealliance/wasmtime/pull/5215).

Fixes #173

- The `ExitCode` property is moved from `TrapException`/`TrapAccessor` to `WasmtimeException`, as the WASI exit code now needs to be retrieved from an `wasmtime_error_t*`.
- The `Frames` property is moved from `TrapException` to `WasmtimeException`, since both errors and traps can have traces now.
- The functionality of passing the original .NET exception as `InnerException` (see #172) is moved from the `TrapException` to the `WasmtimeException`, since host-created traps (with `wasmtime_trap_new()`) are now returned as error (`wasmtime_error_t*`).
  (While theoretically it could happen that the returned `wasmtime_error_t*` has actually another cause than the error returned at the host-to-wasm transition, I assume this won't be the case, similar as to a .NET exception that would just bubble up the stack once thrown.)

Currently, tests that are still failing are the following:
- `FuelConsumptionTests`: Waiting for bytecodealliance/wasmtime#5230


peterhuene:
It appears that `OutOfFuel` trap code is missing from the C API, which is why the fuel tests I think are crashing.

peterhuene:
Fix upstream at https://github.com/bytecodealliance/wasmtime/pull/5230.

kpreisser:
I added the `OutOfFuel` trap code and fixed the definition of `wasmtime_trap_code` (see https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-h84q-m8rr-3v9q) , so once bytecodealliance/wasmtime#5230 is merged, I think this should be complete :slightly_smiling_face:

peterhuene (approved):

peterhuene:
Thanks! Now that it has merged, once the dev artifact is published with that commit, I'll restart CI for this PR and merge.

kpreisser:
Sorry, I forgot to update the exception messages in the `FuelConsumptionTests` (checked with a locally built `Wasmtime`).

peterhuene (dismissed):

peterhuene:
It looks like the CI is now failing for a legitimate reason in the examples, as we're currently expecting a `TrapException` and not `WasmtimeException`.  The former seems correct to me, though?

kpreisser:
Sorry, I hadn't thought of checking the examples. Wasmtime now seems to return the `not enough fuel remaining in store` message as `wasmtime_error_t*` instead of `wasm_trap_t*`, so the example should probably be updated to catch a `WasmtimeException`, similar to the change in the `FuelConsumptionTests`:
https://github.com/bytecodealliance/wasmtime-dotnet/blob/b7dbedf0589908980d5bf3b2a68ac0e7a4468790/tests/FuelConsumptionTests.cs#L85-L89

(Note however, that the `all fuel consumed by WebAssembly` message is still returned as `wasm_trap_t*`, with the new `OutOfFuel` code.)

I can take a look at this tomorrow, unless you beat me to it :slightly_smiling_face:

peterhuene:
I'll push up the fix.

peterhuene (approved):

peterhuene (approved):

peterhuene:
Thanks again for fixing this!

PR #186, opened 2022-11-12
Use unchecked function callbacks for better performance

kpreisser:
Hi,

this PR implements **unchecked** function variants (https://github.com/bytecodealliance/wasmtime/pull/3350) for the generic callback overloads added by #163, which improves performance.

This adds a `ValueRaw` struct that maps to [`wasmtime_val_raw_t`](https://github.com/bytecodealliance/wasmtime/blob/95ca72a37ae17d33d6fa727150dab5a49011b555/crates/c-api/include/wasmtime/val.h#L157-L192) and an `IValueRawConverter<T>` interface similar to the existing ones, which are then used by the unchecked callback functions.

Since the .NET side knows the exact parameter and result types used by the callback, this should be safe. (However I'm not very familiar with wasmtime's resource management, so I would appreciate if you can review especially the handling in the `FuncRefValueRawConverter` and `GenericValueRawConverter`.)

This also improves the NRT annotations for callback delegates (note that for `<auto-generated>` files, we must explicitely add `#nullable enable` even if the project already enables it), and fixes a regression that prevented to define callbacks using an interface as generic type parameter (because that is currently [not allowed by `ValueBox.Converter<T>()`](https://github.com/bytecodealliance/wasmtime-dotnet/blob/44ca7ee4f7bc963e26a460517bd55dae210a3860/src/ValueBox.cs#L209-L214), but it [is allowed by `Value.TryGetKind()`](https://github.com/bytecodealliance/wasmtime-dotnet/blob/44ca7ee4f7bc963e26a460517bd55dae210a3860/src/Value.cs#L210-L214)).

Additionally, callbacks that use neither `Caller` nor `Function` as parameters are now **allocation-free** when being called, since the `Caller` instance will no longer be created as it isn't needed in these cases.

When trying to run the benchmarks from #163 (but using 6 instead of 5 iterations) with .NET 6.0.11 on Windows 10 (Build 19044) x64, I get the following results:

Benchmark 1 (`Action<int>`):

Without this PR:
```
Elapsed: 00:00:00.1257743
Elapsed: 00:00:00.1165437
Elapsed: 00:00:00.1098291
Elapsed: 00:00:00.1045918
Elapsed: 00:00:00.1042576
Elapsed: 00:00:00.1047506
```
With this PR:
```
Elapsed: 00:00:00.0524017
Elapsed: 00:00:00.0489112
Elapsed: 00:00:00.0494636
Elapsed: 00:00:00.0482166
Elapsed: 00:00:00.0423121
Elapsed: 00:00:00.0429405
```

Benchmark 2 (`Action<int, float, long>`):

Without this PR:
```
Elapsed: 00:00:00.2216621
Elapsed: 00:00:00.1976731
Elapsed: 00:00:00.1968243
Elapsed: 00:00:00.1969110
Elapsed: 00:00:00.1974007
Elapsed: 00:00:00.1961490
```
With this PR:
```
Elapsed: 00:00:00.0652740
Elapsed: 00:00:00.0631706
Elapsed: 00:00:00.0615255
Elapsed: 00:00:00.0497510
Elapsed: 00:00:00.0505233
Elapsed: 00:00:00.0497147
```

Benchmark 3 (`Func<int, float, long, ValueTuple<int, int, long>>`):

Without this PR:
```
Elapsed: 00:00:00.4356896
Elapsed: 00:00:00.3905873
Elapsed: 00:00:00.3896729
Elapsed: 00:00:00.3901601
Elapsed: 00:00:00.3902660
Elapsed: 00:00:00.3917198
```
With this PR :
```
Elapsed: 00:00:00.0881153
Elapsed: 00:00:00.0869675
Elapsed: 00:00:00.0849886
Elapsed: 00:00:00.0742805
Elapsed: 00:00:00.0611120
Elapsed: 00:00:00.0619580
```

With the last benchmark, this seems to be roughly a **6x** improvement comparing to the current state (without this PR).

What do you think?

Thanks!


src/ValueRaw.cs
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        public int Unbox(IStore store, in ValueRaw valueRaw)
+        {
+            return valueRaw.i32;
+        }
+
+        public void Box(IStore store, int value, ref ValueRaw valueRaw)
+        {
+            valueRaw.i32 = value;
```
kpreisser, line 93: It seems the corresponding Rust code of this structure (`ValRaw`) [always initializes at least the first 64 bit when setting a value](https://github.com/bytecodealliance/wasmtime/blob/7a6fbe08987a7cb43c7e2a78ba5482bec8ab3c2a/crates/runtime/src/vmcontext.rs#L982-L993):
```rs
impl ValRaw {
    /// Creates a WebAssembly `i32` value
    #[inline]
    pub fn i32(i: i32) -> ValRaw {
        // Note that this is intentionally not setting the `i32` field, instead
        // setting the `i64` field with a zero-extended version of `i`. For more
        // information on this see the comments on `Lower for Result` in the
        // `wasmtime` crate. Otherwise though all `ValRaw` constructors are
        // otherwise constrained to guarantee that the initial 64-bits are
        // always initialized.
        ValRaw::u64((i as u32).into())
    }
```

However, this is not mentioned in the C header file for `wasmtime_val_raw_t`.

Do you know whether this also applies for the values in the C API, i.e. whether we would also need to set the first 64-bit, rather than only the first 32 bits for 32-bit values (so that this case, the next 32 bits will be set to zero)?

kpreisser: I updated the code to always initialize the first 64-bit, so that we match the behavior of the Rust code, as this change doesn't seem to have effects on the performance (at least on 64-bit systems).

Additionally I added a check to ensure we throw a `PlatformNotSupportedException` on Big Endian systems when retrieving a converter, since we would need to implement separate converters for that, which implement switching endianness (as the values in the structure are always in little endian, if I understand the code comments/documentation correctly).

kpreisser: > as the values in the structure are always in little endian, if I understand the code comments/documentation correctly

It seems bytecodealliance/wasmtime#4035 mistakenly added the notes about the little-endian storage to `wasmtime_valunion` instead of `wasmtime_val_raw`; I've opened bytecodealliance/wasmtime#5303 for that.

peterhuene (approved):
@kpreisser thanks for all this work! I paid close attention to the externref/funcref raw value conversion and it seems correct to me.

PR #189, opened 2022-11-23
Fixing Minor Code Quality Issues

martindevans:
Cleaned up some minor issues which were triggering warnings for me:
 - Redundant nullability
 - Redundant `unsafe` contexts
 - Redundant `ToString` calls

Some other cleanup opportunities I did not include in this PR:

#### `public static unsafe extern IntPtr wasmtime_caller_context(IntPtr caller);`
This method (and many others like it) does not need `unsafe`. Should I remove all of them? I've avoided touching anything marked `extern` so far.

#### `checked((int)name->size)` (`size` is `nuint`)
There are many checked conversions from `nuint` to `int`. R# warns: `﻿Checked context is redundant: no operators or conversions with overflow checks`. I'm pretty sure this is an incorrect warning (e.g. see [this test here](https://dotnetfiddle.net/aamMPY)), but I just wanted to double check that before reporting it to Jetbrains.

### DIFF
--- a/src/Caller.cs
+++ b/src/Caller.cs
@@ -134,7 +134,7 @@ namespace Wasmtime
             public static unsafe extern bool wasmtime_caller_export_get(IntPtr caller, byte* name, UIntPtr len, out Extern item);
 
             [DllImport(Engine.LibraryName)]
-            public static unsafe extern IntPtr wasmtime_caller_context(IntPtr caller);
+            public static extern IntPtr wasmtime_caller_context(IntPtr caller);
         }
 
         private IntPtr handle;
--- a/src/Export.cs
+++ b/src/Export.cs
@@ -20,32 +20,29 @@ namespace Wasmtime
             var exports = new Export[(int)this.size];
             for (int i = 0; i < (int)this.size; ++i)
             {
-                unsafe
+                var exportType = this.data[i];
+                var externType = Native.wasm_exporttype_type(exportType);
+
+                switch ((WasmExternKind)Native.wasm_externtype_kind(externType))
                 {
-                    var exportType = this.data[i];
-                    var externType = Native.wasm_exporttype_type(exportType);
+                    case WasmExternKind.Func:
+                        exports[i] = new FunctionExport(exportType, externType);
+                        break;
 
-                    switch ((WasmExternKind)Native.wasm_externtype_kind(externType))
-                    {
-                        case WasmExternKind.Func:
-                            exports[i] = new FunctionExport(exportType, externType);
-                            break;
+                    case WasmExternKind.Global:
+                        exports[i] = new GlobalExport(exportType, externType);
+                        break;
 
-                        case WasmExternKind.Global:
-                            exports[i] = new GlobalExport(exportType, externType);
-                            break;
+                    case WasmExternKind.Table:
+                        exports[i] = new TableExport(exportType, externType);
+                        break;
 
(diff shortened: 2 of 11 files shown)


kpreisser:
Hi,

> #### `checked((int)name->size)` (`size` is `nuint`)
> 
> There are many checked conversions from `nuint` to `int`. R# warns: `﻿Checked context is redundant: no operators or conversions with overflow checks`. I'm pretty sure this is an incorrect warning (e.g. see [this test here](https://dotnetfiddle.net/aamMPY)), but I just wanted to double check that before reporting it to Jetbrains.


I agree this seems to be an incorrect warning, because for `nint` and `nuint` (that are stored as `IntPtr` and `UIntPtr`), the `checked` context is respected by the C# compiler when emitting the conversion in the same way as for other integer types (that don't have user-defined cast operators); without the `checked(...)` context, no `OverflowException` would be thrown when the `nuint` value doesn't fit into an `int` value.

This is in contrast to `UIntPtr` and `IntPtr`, where the user-defined cast operators on these types always threw an `OverflowException` in such a case before .NET 7. Starting with .NET 7, user-defined operators can be [defined as checked and as unchecked version](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/arithmetic-operators#user-defined-checked-operators), which is also done by `UIntPtr` and `IntPtr`. For example (on 64-bit):

```cs
IntPtr x = unchecked((nint)uint.MaxValue);
Console.WriteLine(x);

int y = unchecked((int)x);
Console.WriteLine(y);

int z = checked((int)x);
Console.WriteLine(z);
```

Output with .NET 6:
```
4294967295
Unhandled exception. System.OverflowException: Arithmetic operation resulted in an overflow.
   at System.IntPtr.op_Explicit(IntPtr value)
   at Program.Main() 
```

Output with .NET 7 (notice the `-1` from the `unchecked((int)x)` expression):
```
4294967295
-1
Unhandled exception. System.OverflowException: Arithmetic operation resulted in an overflow.
   at Program.Main() 
```

However, when changing the type of `x` from `IntPtr` to `nint`, the output from .NET 6 will be the same as for .NET 7, because then it no longer uses the user-defined `explicit operator int(IntPtr value)`.

Note, that e.g. `UIntPtr` only defines a cast to `uint` but not `int`, so when casting the `UIntPtr` to `int`, it seems to first invoke the user-defined `explicit operator uint(UIntPtr value)`, and then does the conversion to `int`, so e.g. the following code doesn't throw an `OverflowException` in the cast for`y` even on .NET 6, because the `UIntPtr` can be converted to `uint` without losing information, and then the conversion to `int` is done with respecting the `unchecked` context:

```cs
UIntPtr x = (nuint)uint.MaxValue;
Console.WriteLine(x);

int y = unchecked((int)x);
Console.WriteLine(y);

int z = checked((int)x);
Console.WriteLine(z);
```

Output with .NET 6:
```
4294967295
-1
Unhandled exception. System.OverflowException: Arithmetic operation resulted in an overflow.
   at Program.Main() 
```

> #### `public static unsafe extern IntPtr wasmtime_caller_context(IntPtr caller);`
> 
> This method (and many others like it) does not need `unsafe`. Should I remove all of them? I've avoided touching anything marked `extern` so far.

👍 I agree the `unsafe` can be removed here because it uses `IntPtr` that doesn't need it.

peterhuene (approved):
Thanks for cleaning this up!

PR #191, opened 2022-11-30
Implement bindings for StoreContext data

### LINKED ISSUE #190: Question: Generic Store
In the Rust Wasimtime crate, the wasmtime Store takes a generic parameter for a struct:
```rust
let some_store = SomeStore::new();
let mut store = Store::new(&engine, some_store);
```
The host can then define callback functions that can directly access this data through the caller:
```rust
linker.func_wrap("","some_function", |mut caller: Caller<'_, SomeStore>| {
    let host = caller
      .data()
      .some_store_property
      .as_ref();
}
```
I see this pattern in a lot of places and can't really figure out the best way to replicate this in dotnet, `StoreContext` seems to be not exposed. I feel like I'm missing something. 

Is there functionality available to support something like this? If not, what would it take to implement it?

Thanks!

Yeseh:
closes #190 

This PR adds functionality to the `Store` and `Caller` that allows users to store arbitrary data in the WASM `Store` and access it using host functions like so:

```csharp
var storeData = "Hello, World";
var engine = new Engine();
var linker = new Linker(engine);
var store = new Store(engine, storeData);

linker.DefineFunction("", "store_data", (Caller caller) => {
    var data = caller.GetData() as string;
    Console.WriteLine(data);
});
```

TODO:

- [x] Add binding for `wasmtime_context_get_data`
- [x] Add binding for `wasmtime_context_set_data`

- [x] Add storedata example
- [x] Add test for SetData()

### DIFF
--- a/src/Caller.cs
+++ b/src/Caller.cs
@@ -128,10 +128,15 @@ namespace Wasmtime
         public ulong GetConsumedFuel() => ((IStore)this).Context.GetConsumedFuel();
 
         /// <summary>
-        /// Gets the data from the Store's Context. 
+        /// Gets the user-defined data from the Store's Context. 
         /// </summary>
-        /// <returns></returns>
+        /// <returns>An object represeting the user defined data from this Store</returns>
         public object? GetData() => ((IStore)this).Context.GetData();
+
+        /// <summary>
+        /// Replaces the user-defined data in the Store's Context 
+        /// </summary>
+        public void SetData(object data) => ((IStore)this).Context.SetData(data);
 
         internal static class Native
         {
--- a/src/Caller.cs
+++ b/src/Caller.cs
@@ -136,7 +136,7 @@ namespace Wasmtime
         /// <summary>
         /// Replaces the user-defined data in the Store's Context 
         /// </summary>
-        public void SetData(object data) => ((IStore)this).Context.SetData(data);
+        public void SetData(object? data) => ((IStore)this).Context.SetData(data);
 
         internal static class Native
         {
--- a/src/Caller.cs
+++ b/src/Caller.cs
@@ -127,6 +127,36 @@ namespace Wasmtime
         /// <returns>Returns the fuel consumed by the executing WebAssembly code or 0 if fuel consumption was not enabled.</returns>
         public ulong GetConsumedFuel() => ((IStore)this).Context.GetConsumedFuel();
 
+        /// <summary>
+        /// Gets the data from the Store's Context. 
+        /// </summary>
+        /// <returns></returns>
+        public object? GetData() => ((IStore)this).Context.GetData();
+        
+        /// <summary>
+        /// Gets the data from the Store's Context and attempts to cast it to the specified type. 
+        /// </summary>
+        /// <returns>The typed store data, or null in case of a failed cast</returns>
+        public T? GetData<T>() 
+            where T: class
+        {
+            var data = this.GetData();
+            if (data is null) { return null; }
+
+            try  
+            { 
+                return (T)data; 
+            }
+            catch (InvalidCastException _) 
+            { 
+                return null; 
+            }
+            catch
+            {
+                throw;
+            }
+        }
+
         internal static class Native
         {
             [DllImport(Engine.LibraryName)]
--- a/src/Caller.cs
+++ b/src/Caller.cs
@@ -132,30 +132,6 @@ namespace Wasmtime
         /// </summary>
         /// <returns></returns>
         public object? GetData() => ((IStore)this).Context.GetData();
-        
-        /// <summary>
-        /// Gets the data from the Store's Context and attempts to cast it to the specified type. 
-        /// </summary>
-        /// <returns>The typed store data, or null in case of a failed cast</returns>
-        public T? GetData<T>() 
-            where T: class
-        {
-            var data = this.GetData();
-            if (data is null) { return null; }
-
-            try  
-            { 
-                return (T)data; 
-            }
-            catch (InvalidCastException _) 
-            { 
-                return null; 
-            }
-            catch
-            {
-                throw;
-            }
-        }
 
         internal static class Native
         {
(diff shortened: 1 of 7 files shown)


src/Caller.cs
```diff
@@ -127,6 +127,36 @@ private IntPtr NativeHandle
...
+                return (T)data; 
+            }
+            catch (InvalidCastException _) 
+            { 
+                return null; 
+            }
+            catch
+            {
+                throw;
+            }
```
kpreisser, line 157: I think you can just use the `as` operator here, to avoid throwing (and catching) the `InvalidCastException`, as it just returns `null` when the value cannot be cast to that type, which is what the above code effectively does:

```suggestion
            return data as T;
```

src/Store.cs
```diff
@@ -90,6 +97,10 @@ private static class Native
 
             [DllImport(Engine.LibraryName)]
             public static extern void wasmtime_context_set_epoch_deadline(IntPtr handle, ulong ticksBeyondCurrent);
+            
+            [DllImport(Engine.LibraryName)]
+            [return: MarshalAs(UnmanagedType.IUnknown)]
+            public static extern object wasmtime_context_get_data(IntPtr handle);
```
kpreisser, line 103: `MarshalAs(UnmanagedType.IUnknown)` is intended for COM interop on Windows, which I don't think is appropriate here. I think you should declare it as returning an `IntPtr`, and then use `GCHandle.FromIntPtr(value).Target!` to get the target object from the `GCHandle`.

See https://github.com/bytecodealliance/wasmtime-dotnet/blob/992efb0baee78bad1b4f9cc0a8a5b363313a63c9/src/Value.cs#L384 for an example.

src/Caller.cs
Yeseh: Wasn't aware that as gives back null, that's handy. I think in that case GetData<T> doesn't do much anymore, so I removed it.

src/Store.cs
Yeseh: Makes sense, fixed

Yeseh:
Thanks for the feedback @kpreisser !

I'll fix the failing test case (dumb mistake) and add an example to the funcrefs later today or tomorrow. Anything else you can think of? Are the two testcases sufficient?

Yeseh:
Added an example and actually working testcases together with the binding for `wasmtime_context_set_data`. Think it's functionally finished right now.

src/Store.cs
```diff
@@ -28,6 +28,29 @@ internal void AddFuel(ulong fuel)
...
+            var data = Native.wasmtime_context_get_data(handle);
+            if (data == IntPtr.Zero)
+            {
+                return null;
+            }
+
+            return GCHandle.FromIntPtr(data).Target;
+        }
+
+        internal void SetData(object data)
```
kpreisser, line 42: If I understand this correctly, when you initially create the `Store` without passing data or passing `null`, but later call `SetData`, then the finalizer would not be called and the `GCHandle` would never be freed, because in `wasmtime_store_new`, `null` would have been passed as finalizer.

(That might be a good candidate for a unit test, e.g. similar to the `ExternRefTests.ItCollectsExternRefs` that checks whether `externref` object instances are freed by the GC after disposing the `Store`.)

I think in order to make this work, we would need to always pass the finalizer to `wasmtime_store_new`, and change the finalizer to something like this (as `IntPtr.Zero` would be an invalid GC handle):
```cs
private static readonly Native.Finalizer Finalizer = (p) =>
{
    if (p != IntPtr.Zero)
    {
        GCHandle.FromIntPtr(p).Free();
    }
};
```

Additionally, `SetData` can then be changed to accept `object?`, and call `GCHandle.Alloc()` only if the new data is not `null`. What do you think?

Yeseh: Yes totally agree, good catch. This way calling `SetData(null)` will do what you'd expect, and clear the existing store data. I Added tests to check if GC properly takes care of unused store data on `SetData()` and `Store.Dispose()`

src/Store.cs
```diff
@@ -28,6 +28,32 @@ internal void AddFuel(ulong fuel)
...
+            if (oldData != IntPtr.Zero) 
+            {
+                GCHandle.FromIntPtr(oldData).Free();
+            }
+            
+            if (data != null)
+            {
+                var newPtr = (IntPtr)GCHandle.Alloc(data);
+                Native.wasmtime_context_set_data(handle, newPtr);
+            }
```
kpreisser, line 54: The call to `wasmtime_context_set_data` needs to be made unconditionally, to ensure we set the store data actually to `null`.
Otherwise, when calling `SetData(null)` and previously there was an object set, we would free the `GCHandle` here, but then the finalizer would attempt to free it again when disposing of the `Store` later (which can cause undefined behavior as the handle is no longer valid).

```suggestion
            var newPtr = IntPtr.Zero;
            if (data != null)
            {
                var newPtr = (IntPtr)GCHandle.Alloc(data);
            }            

            Native.wasmtime_context_set_data(handle, newPtr);
```

kpreisser: Additionally, to be on the safe side if allocating a new `GCHandle` (or calling `wasmtime_context_set_data`) should ever fail, I think it might be a good idea to free the old handle only after calling `wasmtime_context_set_data` succeeded, i.e. move 
```cs
            if (oldData != IntPtr.Zero) 
            {
                GCHandle.FromIntPtr(oldData).Free();
            }
```
after the call to `wasmtime_context_set_data`, to prevent the situation that we free the `GCHandle` but it's still set as store data, so that the finalizer would then free it again.
Thanks!

Yeseh: Done! Thanks a lot for being patient and helping me figure out a bunch of new concepts, your explanations were very helpful.

peterhuene:
@Yeseh thanks again for implementing this!

Yeseh:
@peterhuene just noticed that we all missed the storedata example being in the root folder of the repo instead of examples/ :)

peterhuene:
Ah, how little one pays attention to a path when there's fun code to review 😅.

Thankfully the repo layout doesn't matter much and we can address it easily with a follow up.

PR #193, opened 2022-12-03
Move allocating the `Caller` into the `try` block (follow-up to #188)

kpreisser:
In #188, I missed that allocating the `Caller` was done outside of the `try` block, and that could potentially throw an `OutOfMemoryException` when the system is low on memory.

This PR moves allocating the `Caller` into the `try` block, to prevent the .NET Runtime on Windows unwinding the stack to the next .NET exception handler in case the allocation threw an OOME (which would be incompatible with Wasmtime) and instead attempts to convert it into a `wasm_trap_t*`.

Thanks!


peterhuene (approved):

peterhuene:
Thanks!

PR #201, opened 2022-12-09
Store the `Store` object itself in the context data

### LINKED ISSUE #200: `Function` instance passed to a callback (as `funcref`) cannot be called outside of the callback
Currently, when a callback gets a `funcref` value, it will be created as a `Function` instance, using the `Caller` as `IStore`. However, the `Caller` is only valid within the callback, so it's not possible to store the `Function` and later call it outside of the callback.

Consider the following code:
```cs
    engine,
    "hello",
    @"
(module 
    (import """" ""store_funcref"" (func $store_funcref (param funcref)))
    (export ""run"" (func $run))
    (elem declare func $getValue)
    (func $getValue (result i32)
        i32.const 123
    )
    (func $run 
        (call $store_funcref (ref.func $getValue))
    )
)");

var storedFuncRef = default(Function);
linker.DefineFunction("", "store_funcref", (Function? funcRef) =>
{
    storedFuncRef = funcRef;
});

run();

var wrappedFuncRef = storedFuncRef!.WrapFunc<int>()!;
int result = wrappedFuncRef();
Console.WriteLine("Result: " + result);
```

**Expected behavior:**

Runs successfully and prints `Result: 123` to the console.

**Actual behavior:**

Throws an `ObjectDisposedException`, because the `Caller` is no longer usable.
```
Unhandled exception. System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'Wasmtime.Caller'.
   at Wasmtime.Caller.get_NativeHandle() in C:\Users\VPC\Desktop\WebAssembly-Test\wasmtime-dotnet\src\Caller.cs:line 92
   at Wasmtime.Caller.Wasmtime.IStore.get_Context() in C:\Users\VPC\Desktop\WebAssembly-Test\wasmtime-dotnet\src\Caller.cs:line 99
   at Wasmtime.Function.<>c__DisplayClass184_0`1.<WrapFunc>b__0() in C:\Users\VPC\Desktop\WebAssembly-Test\wasmtime-dotnet\src\Function.Wrap.cs:line 841
   at Program.<<Main>$>g__TestCallbackStoreFuncRef|0_9() in C:\Users\VPC\Desktop\WebAssembly-Test\TestWasmtime\Program.cs:line 669
   at Program.<Main>$(String[] args) in C:\Users\VPC\Desktop\WebAssembly-Test\TestWasmtime\Program.cs:line 687
```

<del>With the work done by #191, I was thinking to leverage the store data feature, and instead of storing the store's `Data` object, store the `Store` object itself in the context data, so that we can retrieve it from the caller and can use it to create `Function` instances that are still usable outside of the callback.</del>

kpreisser:
Hi,

with this PR, we leverage the store context data work introduced with #191 in order to store a reference to the `Store` itself, so that we can retrieve the originating `Store` from a `Caller`.

When we create a `Function` instance from a callback from a `funcref` value, we will then use the `Store` instead of the `Caller` to instantiate the `Function`, so that it is also callable outside of the callback.

This also allows some code clean-up e.g. in the `IValueRaw` converters, since we can now unconditionally pass a `Store` that doesn't need to be allocated.

Fixes #200

What do you think?

Thanks!


peterhuene:
So I have a few problems with this approach, chiefly that it would inherently introduce a circular reference between the .NET and the native store representations making it much easier to leak a `Store` and by extension everything used in that store.

Currently if you forget to dispose a .NET `Store` and it becomes eligible for collection because it isn't reachable from any .NET roots, it will get collected and the contained `Handle` to the native store object will be freed during finalization, thus "dropping" the native store and releasing the root on the data object (if there is one).

With these changes, the native store always holds a root to the `Store` and the `Store` holds a handle to the native store, meaning it'll never be collected without an explicit `Dispose` call to free the root held by the native store.

While it _is_ possible to reference a `Store` from a store data object *today* and thus introduce the same circular reference, we should probably add a "don't do that unless you really know what you're doing" documentation warning somewhere.

Additionally, I'm not convinced we need to fix #200. Linker functions should not capture anything related to caller context and should manipulate any state that should outlive the call via `Caller.GetData`. While in Rust it's easy to enforce this invariant thanks to its type system and the compiler's borrow-checker, it's unfortunately trivial to violate in a GC language like C# with zero warning.

There will likely be other things which violate an invariant that the underlying Rust API holds that we will likely just have to say "don't do it that way, do it this way instead" to users.

peterhuene:
The "proper" way to fix this, in my opinion, is to go back to passing in the store context to use when invoking a function, like in the Rust API. Thus `Function` would not store the context association meaning that it could safely outlive the caller context to be called with another reference to the same store.

kpreisser:
Hi @peterhuene,
thanks for your reply and review. Yes, I agree in particular the circular dependency would be a bad thing to introduce.
I will look into possible alternatives for fixing the issue.
Thanks!

PR #202, opened 2022-12-13
Reduced allocations

martindevans:
- Using `Array.Empty` where possible
 - Get UTF8 bytes into a `Memory<byte>` rented from the shared memory pool instead of a `byte[]`
 - Replaced some LINQ with a `for` loop
 - Modified some private helper functions to return more appropriate types based on their ultimate uses
 - Removed `TypeHandle` used in `Memory` - it was only used in two places with trivial lifetimes.


src/Linker.DefineFunction.tt
```diff
@@ -90,8 +96,14 @@ foreach (var (hasCaller, resultCount, parameterCount, methodGenerics, delegateTy
...
 
-                var moduleBytes = Encoding.UTF8.GetBytes(module);
-                var nameBytes = Encoding.UTF8.GetBytes(name);
+                var moduleLength = Encoding.UTF8.GetByteCount(module);
+                Span<byte> moduleBytes = stackalloc byte[moduleLength];
+                Encoding.UTF8.GetBytes(module, moduleBytes);
+
+                var nameLength = Encoding.UTF8.GetByteCount(name);
+                Span<byte> nameBytes = stackalloc byte[nameLength];
+                Encoding.UTF8.GetBytes(name, nameBytes);
```
kpreisser, line 105: Maybe guard against stack overflows (that would terminate the process) in case the user supplies a large string for some reason as `module` or `name`?
```suggestion
                const int stackAllocThreshold = 256;
                
                var moduleLength = Encoding.UTF8.GetByteCount(module);
                Span<byte> moduleBytes = moduleLength <= stackAllocThreshold ? stackalloc byte[moduleLength] : new byte[moduleLength];
                Encoding.UTF8.GetBytes(module, moduleBytes);

                var nameLength = Encoding.UTF8.GetByteCount(name);
                Span<byte> nameBytes = nameLength <= stackAllocThreshold ? stackalloc byte[nameLength] : new byte[nameLength];
                Encoding.UTF8.GetBytes(name, nameBytes);
```

Alternatively, e.g. in the .NET Runtime/BCL, a common pattern for this is to use `ArrayPool<T>.Shared` to rent an array if the length exceeds the stackalloc threshold, and return it later, which can also minimize allocations in that case.
See this example:
https://github.com/dotnet/runtime/blob/3689fbec921418e496962dc0ee252bdc9eafa3de/src/libraries/System.Net.Http/src/System/Net/Http/MultipartContent.cs#L653-L673

martindevans: I've rewritten it to use the shared `MemoryPool<byte>` instead of stack allocations.

kpreisser: Thanks!
However, note that, unfortunately, AFAIK using `MemoryPool<T>.Shared.Rent` will always cause an allocation of at least one object (in contrast to `ArrayPool<T>.Shared.Rent`), in this case the `IMemoryOwner<T>`. See these code locations in the runtime:
https://github.com/dotnet/runtime/blob/322f20cda9673a3872902ed3fb2ddfbcff97c007/src/libraries/System.Memory/src/System/Buffers/ArrayMemoryPool.cs#L21
https://github.com/dotnet/runtime/blob/322f20cda9673a3872902ed3fb2ddfbcff97c007/src/libraries/System.Memory/src/System/Buffers/ArrayMemoryPool.ArrayMemoryPoolBuffer.cs#L8-L15

I think for best efficiency (completely avoid heap allocations in the most common case, when the `module` and `name` strings are not very long), it might make sense to apply the implementation from the `MultipartContent` example, which uses `stackalloc byte[...]` for strings shorter than the threshold, and uses `ArrayPool<byte>.Shared.Rent(...)` for longer strings.

What do you think?

src/Linker.DefineFunction.tt
```diff
@@ -90,26 +97,29 @@ foreach (var (hasCaller, resultCount, parameterCount, methodGenerics, delegateTy
...
+                    handle,
+                    (byte*)modulePin.Pointer,
+                    (nuint)moduleBytesLength,
+                    (byte*)namePin.Pointer,
+                    (nuint)nameBytesLength,
+                    funcType,
+                    func,
+                    GCHandle.ToIntPtr(GCHandle.Alloc(func)),
+                    Function.Finalizer
+                );
```
kpreisser, line 118: > I've rewritten it to use the shared `MemoryPool<byte>` instead of stack allocations.

Thanks!
However, note that, unfortunately, AFAIK using `MemoryPool<T>.Shared.Rent` will always cause an allocation of at least one object (in contrast to `ArrayPool<T>.Shared.Rent`), in this case the `IMemoryOwner<T>`. See these code locations in the runtime:
https://github.com/dotnet/runtime/blob/322f20cda9673a3872902ed3fb2ddfbcff97c007/src/libraries/System.Memory/src/System/Buffers/ArrayMemoryPool.cs#L21
https://github.com/dotnet/runtime/blob/322f20cda9673a3872902ed3fb2ddfbcff97c007/src/libraries/System.Memory/src/System/Buffers/ArrayMemoryPool.ArrayMemoryPoolBuffer.cs#L8-L15

I think for best efficiency (completely avoid heap allocations in the most common case, when the `module` and `name` strings are not very long), it might make sense to apply the implementation from the `MultipartContent` example, which uses `stackalloc byte[...]` for strings shorter than the threshold, and uses `ArrayPool<byte>.Shared.Rent(...)` for longer strings (although the code is a bit more verbose).

What do you think?

```suggestion
                const int StackallocThreshold = 256;

                byte[]? moduleBytesBuffer = null;
                var moduleLength = Encoding.UTF8.GetByteCount(module);
                Span<byte> moduleBytes = moduleLength <= StackallocThreshold ? stackalloc byte[moduleLength] : (moduleBytesBuffer = ArrayPool<byte>.Shared.Rent(moduleLength))[..moduleLength];
                Encoding.UTF8.GetBytes(module, moduleBytes);

                byte[]? nameBytesBuffer = null;
                var nameLength = Encoding.UTF8.GetByteCount(name);
                Span<byte> nameBytes = nameLength <= StackallocThreshold ? stackalloc byte[nameLength] : (nameBytesBuffer = ArrayPool<byte>.Shared.Rent(nameLength))[..nameLength];
                Encoding.UTF8.GetBytes(name, nameBytes);

                try
                {
                    fixed (byte* modulePtr = moduleBytes, namePtr = nameBytes)
                    {
                        var error = Native.wasmtime_linker_define_func_unchecked(
                            handle,
                            modulePtr,
                            (nuint)moduleBytes.Length,
                            namePtr,
                            (nuint)nameBytes.Length,
                            funcType,
                            func,
                            GCHandle.ToIntPtr(GCHandle.Alloc(func)),
                            Function.Finalizer
                        );

                        if (error != IntPtr.Zero)
                        {
                            throw WasmtimeException.FromOwnedError(error);
                        }
                    }
                }
                finally
                {
                    if (moduleBytesBuffer is not null)
                    {
                        ArrayPool<byte>.Shared.Return(moduleBytesBuffer);
                    }
                    if (nameBytesBuffer is not null)
                    {
                        ArrayPool<byte>.Shared.Return(nameBytesBuffer);
                    }
                }
```

martindevans: You're absolutely right, not sure how I missed that MemoryPool would allocate that handle 🤦‍♂️

I only used `MemoryPool` instead of `ArrayPool` out of neatness - it looked cleaner to have a `using` instead of wrapping everything in try/finally!

src/Function.FromCallback.tt
```diff
@@ -57,7 +57,12 @@ foreach (var (hasCaller, resultCount, parameterCount, methodGenerics, delegateTy
             var parameterKinds = new List<ValueKind>();
             var resultKinds = new List<ValueKind>();
 
+            <# if (callbackParameterTypeExpressions.Length == 0) { #>
+            var callbackParameterTypes = Array.Empty<Type>();
+            <# } else { #>
             var callbackParameterTypes = new Type[] { <#= callbackParameterTypeExpressions #>};
+            <# } #>
```
kpreisser, line 64: nit: Note that this syntax causes an additional indentation for the line in the generated `.cs` file:
```diff
-            var callbackParameterTypes = new Type[] { typeof(T), };
+                        var callbackParameterTypes = new Type[] { typeof(T), };
```

kpreisser (approved):

src/Function.FromCallback.tt
martindevans: Thanks, fixed 👍

martindevans:
CI failure due to an issue downloading a dependency:
> Response status code does not indicate success: 503 (Egress is over the account limit.).

martindevans:
Rebased onto master to fixed merge conflict.

src/Export.cs
```diff
@@ -17,6 +17,9 @@ public void Dispose()
 
         public Export[] ToExportArray()
         {
+            if (size == UIntPtr.Zero)
+                return Array.Empty<Export>();
```
peterhuene, line 21:
```suggestion
            if (size == UIntPtr.Zero)
            {
                return Array.Empty<Export>();
            }
```

src/Import.cs
```diff
@@ -30,6 +30,9 @@ public void Dispose()
 
         public Import[] ToImportArray()
         {
+            if (size == UIntPtr.Zero)
+                return Array.Empty<Import>();
```
peterhuene, line 34:
```suggestion
            if (size == UIntPtr.Zero)
            {
                return Array.Empty<Import>();
            }
```

peterhuene (approved):
Looks good, just two formatting nits 👍

martindevans:
Comitted those two suggestions. Thanks for the review 👍

PR #203, opened 2022-12-14
Fix Inconsistent Line Endings

martindevans:
Since the T4 templates were introduced I've been getting warnings from Visual Studio/Unity about inconsistent line endings. This seems to be because the `StringBuilder` in `FunctionCallbackOverloadTemplates.t4` is using `\r\n` (the correct line ending for Windows) but the t4 file itself is written with `\n` line endings.

For now I've modified the template to replace all generated `\r\n` with `\n` for consistency. But I'm not sure that's ultimately the right approach, can anyone suggest any better alternatives (e.g. can t4 be made to normalize things)?

### DIFF
--- a/src/FunctionCallbackOverloadTemplates.t4
+++ b/src/FunctionCallbackOverloadTemplates.t4
@@ -1,5 +1,5 @@
 ﻿<#
-// This file contains the functions used in Function.FromCallback.ts and Linker.DefineFunction.ts
+// This file contains the functions used in Function.FromCallback.tt and Linker.DefineFunction.tt
 // in order to generate overloads with different parameter and result types for Action<...>/Func<...>.
 
 // Suppress the "not used" warning when this file is referenced but the function is not called.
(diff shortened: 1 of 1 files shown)


kpreisser (changes requested):
Git has an [option](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings) to configure how line-endings should be handled, where the default behavior (`auto`) on Windows is to check out text files with an CR-LF ending, but to store them internally with a LF ending when committing them.

On my system, when I clone the repo, all the checked-out `.cs` and `.tt` files have a CR-LF encoding, and when running the build to generate the files from the `.tt` files, the generated files still have the CR-LF encoding as that is the default line ending (`\r\n`) in .NET on Windows (`Environment.NewLine`), whereas on Linux, the .NET line ending would be LF (`\n`), and Git on Linux would also check out text files with an `LF` ending with the `auto` setting.

Ironically, if I checkout your PR branch (`fix_line_endings`) on Windows using default Git settings and build the project, then I get now warnings about inconsistent line endings (due to some code lines of the generated files now using `\n` instead of `\r\n`).

Are you using a setting that causes text files to be checked out with `LF` ( `\n`) line endings on Windows?

I think the "correct" fix would probably be to create a `.gitattributes` file that specifies the line ending setting for this repo, to avoid such issues when you have a different global setting. What do you think?

Thanks!

martindevans:
I'm not aware that I've changed that setting but it looks like it was set incorrectly for me. Thanks for pointing me in the rught direction 👍

PR #205, opened 2022-12-15
Yet more tests

martindevans:
Added a number of tests.

In particular found one test which would crash the entire test runner (seems to be due to an access violation exception):

```csharp
[Fact]
public void ItFailsWithInvalidMutability()
{
    var act = () => new Global(Store, ValueKind.Int64, 0, (Mutability)13);
    act.Should().Throw<InvalidOperationException>();
}
```

This was caused by two things:
 - Immediate problem was a missing check if the returned value is `IntPtr.zero` in this one particular case (added that check).
 - Casting numbers into Mutability should not be possible. Replaced `Mutability` with a struct which has two predefined values (Mutable and Immutable). This should be _almost_ source compatible with the old enum - the only thing it does not allow is casting numbers into `Mutability`.

### DIFF
--- a/src/Global.cs
+++ b/src/Global.cs
@@ -25,7 +25,9 @@ namespace Wasmtime
         internal Mutability(byte value)
         {
             if (value > 1)
+            {
                 throw new ArgumentOutOfRangeException($"Invalid Mutability value `{value}`");
+            }
 
             Value = value;
         }
--- a/src/Global.cs
+++ b/src/Global.cs
@@ -7,16 +7,68 @@ namespace Wasmtime
     /// <summary>
     /// Represents the mutability of a WebAssembly global value.
     /// </summary>
-    public enum Mutability : byte
+    public readonly struct Mutability
+        : IEquatable<Mutability>
     {
         /// <summary>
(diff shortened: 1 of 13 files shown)


src/Global.cs
```diff
@@ -7,16 +7,68 @@ namespace Wasmtime
...
         /// </summary>
-        Mutable,
+        public static readonly Mutability Mutable = new(1);
+
+        internal readonly byte Value;
+
+        internal Mutability(byte value)
+        {
+            if (value > 1)
+                throw new ArgumentOutOfRangeException($"Invalid Mutability value `{value}`");
```
peterhuene, line 28:
```suggestion
            {
                throw new ArgumentOutOfRangeException($"Invalid Mutability value `{value}`");
            }
```

peterhuene (approved):
Thanks for this and apologies for the long time getting around to it (vacation + focusing on other things the past few weeks).

Just one little nit formatting comment, otherwise I think the code changes make sense and the test changes are welcomed.

martindevans:
Thanks for the review, I hope you had a good holiday :)

PR #206, opened 2022-12-17
Keep the `Store` alive until its `StoreContext` is no longer used

kpreisser:
Keep the `Store` alive until its `StoreContext` is no longer used.

As noted in https://github.com/bytecodealliance/wasmtime-dotnet/issues/200#issuecomment-1355191511, this is required (to handle the case when you forget to dispose the `Store`) as otherwise it could theoretically happen that the `Store` handle is deleted (with `wasmtime_store_delete`) in the GC finalizer thread even when at the same time a native call using the `StoreContext` handle is still executing in another thread, in case the `Store` object is no longer reachable.

For native methods taking a handle parameter that is passed as `SafeHandle`, this is not required as the `SafeHandle` is already kept alive during the call.
An exception is if you use `SafeHandle.DangerousGetHandle()` to retrieve the handle as pointer value and pass it; in that case you must also keep the `SafeHandle` alive.

This commit also fixes an instance of the above noted issue in `Engine.IncrementEpoch()`, where `wasmtime_engine_increment_epoch` was declared as taking an `IntPtr` handle, and `Engine.IncrementEpoch()` used `SafeHandle.DangerousGetHandle()` without keeping the SafeHandle alive (this appears to have been introduced with #118).

Note that I also added the `GC.KeepAlive(store)` if the store is used in method calls after that, because otherwise it is not guaranteed that the following call will actually keep the store alive (the call could be inlined, and if then doesn't read the `Store`, it would still be eligible for GC); additionally, this would cause a "code debt" if you edit the code in the future, as e.g. when you would change/remove the code after the call using the `StoreContext`, you would have to remember to add the `GC.KeepAlive(store)` call.

---

**Note:** A general question that came to my mind, is whether `Wasmtime` actually has a thread-safe resource management, i.e. whether it supports that one thread may be freeing a resource, while at the same time another thread is calling a native method that uses another value that internally might contain a reference to the resource freed by the first thread.

For example, there is the `Engine` object (`wasm_engine_t`) that you need to create first, and then you can create a `Store` (`wasmtime_store_t*`) from the `Engine` (as `wasmtime_store_new()` takes a `wasm_engine_t*` parameter), from which I understand that internally, the returned `wasmtime_store_t` may contain a reference to the `wasm_engine_t`.
(So, if you call `wasm_engine_delete` to delete the engine while a store still exists, this will not actually free the `wasm_engine_t` but just decrements its reference count. If you later delete the `wasmtime_store_t`, then it and the referenced engine will actually be freed.)

However, if in .NET code you create an `Engine`, and then create a `Store` from the `Engine` and then throw the `Engine` away (and you forget to use a `using` block or to call `Engine.Dispose()` afterwards), and later you call `Store.Dispose()`, it can happen that `wasm_engine_delete()` is called from the GC finalizer thread, while at the same time your main thread is calling `wasmtime_store_delete()`. Is this supported by `Wasmtime`?

(If not, them I'm wondering whether the apporach of releasing handles in finalizers can actually be supported at all by `wasmtime-dotnet`.)

What do you think?

Thanks!

### DIFF
--- a/src/Function.FromCallback.cs
+++ b/src/Function.FromCallback.cs
@@ -81,6 +81,8 @@ namespace Wasmtime
                     out var externFunc
                 );
 
+                GC.KeepAlive(store);
+
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
         }
@@ -153,6 +155,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -229,6 +233,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -308,6 +314,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -390,6 +398,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -475,6 +485,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -563,6 +575,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -654,6 +668,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -748,6 +764,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -845,6 +863,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -945,6 +965,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -1048,6 +1070,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -1155,6 +1179,8 @@ namespace Wasmtime
                     out var externFunc
                 );
 
+                GC.KeepAlive(store);
+
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
         }
@@ -1228,6 +1254,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -1304,6 +1332,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -1383,6 +1413,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -1465,6 +1497,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -1550,6 +1584,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -1638,6 +1674,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
@@ -1729,6 +1767,8 @@ namespace Wasmtime
                     Finalizer,
                     out var externFunc
                 );
+
+                GC.KeepAlive(store);
 
                 return new Function(store, externFunc, parameterKinds, resultKinds);
             }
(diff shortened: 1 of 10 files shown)


peterhuene:
Apologies on the delay for reviewing this as I was on vacation in December and I've been focusing a lot on tooling unrelated to the .NET bindings lately.

I'll read up on the context and review the changes shortly.

peterhuene:
> A general question that came to my mind, is whether Wasmtime actually has a thread-safe resource management, i.e. whether it supports that one thread may be freeing a resource, while at the same time another thread is calling a native method that uses another value that internally might contain a reference to the resource freed by the first thread.

Wasmtime follows Rust's threading model where `Store<T>` is `Send` if `T` is `Send` and `Sync` if `T` is `Sync`, meaning that the state stored determines how the type can be sent and shared between threads.

Unfortunately we can't easily maintain the same invariants in .NET as we can in Rust, so it boils down to:

peterhuene:
> A general question that came to my mind, is whether Wasmtime actually has a thread-safe resource management, i.e. whether it supports that one thread may be freeing a resource, while at the same time another thread is calling a native method that uses another value that internally might contain a reference to the resource freed by the first thread.

Wasmtime follows Rust's threading model where `Store<T>` (here `T` represents the user context data) is `Send` if `T` is `Send` and `Sync` if `T` is `Sync`, meaning that the context type determines whether or not the store can be sent and synchronized between threads.

It further relies on Rust to maintain a guarantee that `Store`-related objects cannot outlive the `Store` itself; this is accomplished by requiring a reference to the store be passed into any operation on a `Store`-related object; thus the Rust compiler can statically verify that a `Store` outlives its use by the `Store`-related objects.
 
Some types, like `Engine` and `Module`, are inherently `Send` and `Sync` and can be used safely from multiple threads; this includes deleting them as references to these are maintained via a synchronized reference count. Basically the _only_ thing the .NET API should be doing from the finalizer thread is decrementing these reference counts.

The real problem with the .NET API is that it doesn't uphold the same guarantees with `Store`:

* There's no way to automatically enforce safety for store context data (i.e. the `T` from above); if users supply their own context data then they must ensure it can be sent between threads if they send the `Store` between threads and synchronized if more than one thread can mutate the context data. This can only be solved with documentation/guidance from the .NET API.

* The .NET API doesn't uphold the same guarantee regarding the _use_ of `Store`-related objects that the Rust API does; `Store`-related objects in the .NET API hold strong references on the `Store` they come from (albeit in a way that makes it not possible to keep around `Caller`-related objects past a host call, currently). The upside of this is that it makes the API more ergonomic in terms of not having to pass a reference to the `Store` for every operation on a `Store`-related object. The downside is that it is possible to explicitly `Dispose` a `Store` while it is still in use by a related object, which is unsafe.

There's two possible solutions in my mind for the second issue:

* We simply document that users of the API must guarantee that `Store` is not disposed until _after_ every `Store`-related object is no longer usable. It's not ideal, but also not the most arduous of requirements given the .NET API is effectively an interop shim. For most use cases, I imagine this is very easy to maintain as one would create a `Store`, interact with it, and then dispose it without involving other threads.

* Go back to explicitly accepting the `Store` to use for operations on `Store`-related objects and no longer keep a strong reference on the `Store` from the related objects. This guarantees that the `Store` remains alive for the duration of the requested operation. While not as ergonomic, this would more closely model how things are done in Wasmtime itself.

I haven't yet looked over the PR, but I am, on the surface, wary of injecting `GC.KeepAlive` in lots of places. I'll have to dig deeper and see what's necessary.

src/Function.FromCallback.cs
```diff
@@ -81,6 +81,8 @@ public static Function FromCallback(IStore store, Action callback)
                     out var externFunc
                 );
 
+                GC.KeepAlive(store);
```
peterhuene, line 84: Shouldn't `store` remain stack-rooted for the duration of this function given it's a parameter to this method and passed as an argument to `Function`'s constructor (which will then store the reference)? On the surface, I don't see how `store` could be GC'd at any point in this method.

I'm not in anyway familiar with the optimizations the JIT compiler can make to shorten the root's lifetime in a way that would impact the correctness of this code, though. 

Would you mind further elaborate on the need for `GC.KeepAlive` here?

peterhuene: I see, perhaps this guards against `Function`'s constructor call getting inlined and the caller discards the return value (at every call site? does the JIT compiler even make that optimization?), thus optimizing away any further use of the `store` parameter such that the lifetime is shortened to just after the `store.Context.handle` access?

But if this happens, it means the caller isn't using the `store` argument after this call either and since it discarded the returned `Function` (which must keep `store` reachable), then no other store-related objects are reachable either.

If that is true, then `store` can be finalized at this `GC.KeepAlive` site without a correctness issue, no?

peterhuene: Except `store.Context.handle` is `IntPtr` and not `Handle`, which isn't being rooted _during_ the call to `wasmtime_func_new`.

Thus any attempts internally in the Wasmtime API to access the `Store` might be unsafe.

Now I see the need for it! Sorry for the manual thought process here.

peterhuene:
I followed up with #200 to say that I think that storing a weak self-reference in the store context data, as you originally described, would be a good path to move forward on for fixing the issue with `Caller` retrieved objects being tied to the lifetime of the call and having the least impact on the current API.

I'm going to move forward with this review to fix the store lifetime issues at interop call sites.

src/Store.cs
```diff
@@ -138,6 +138,12 @@ public interface IStore
         /// <summary>
         /// Gets the context of the store.
         /// </summary>
+        /// <remarks>
+        /// Note: Generally, you must keep the <see cref="IStore"/> alive (by using
+        /// <see cref="GC.KeepAlive(object)"/>) until the <see cref="StoreContext"/> is no longer
+        /// used, to prevent the the Store.Handle finalizer from prematurely deleting the handle 
+        /// in the GC finalizer thread while the <see cref="StoreContext"/> is still in use.
+        /// </remarks>
         StoreContext Context { get; }
```
peterhuene, line 147: Thanks for adding this comment about safety!

As I mentioned in #200, I think it might be for the best that we eliminate `IStore` since it exposes this unsafe behavior unnecessarily. If we do, we should definitely keep this comment around such that wherever we construct a `StoreContext` the caller must know to keep the store alive for the duration of it being used.

peterhuene (approved):
This looks good and thanks for being so thorough!

src/Function.FromCallback.cs
kpreisser: Thanks! Yes, the issue is that the `Store.Handle` instance isn't rooted during the `wasmtime_func_new` call, which can mean that the GC already collects the `Function` instance together with referenced the `Store` instance and thus the Store's `Handle` instance (if the caller doesn't use the returned `Function`) and calls its finalizer, and so it could happen that `wasmtime_store_delete` has already been called or is currently being called (from the GC finalizer thread) when the current thread enters the native `wasmtime_func_new` where the store context obtained from the store is passed as `IntPtr`.

To prevent this, `GC.KeepAlive()` can used, which does nothing by itself, but instructs the runtime to treat the specified reference as being reachable until after that call. This pattern is also used by the .NET runtime libraries itself, for example in the `System.Windows.Forms.Control` class (there, it prevents the Windows `hWnd` from being freed): https://github.com/dotnet/winforms/blob/9184e02413ee442f30d2548db84088bb7336edcc/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs#L1535-L1536

Alternatively (at least when we actually have a `Store`, not an `IStore`), instead of using `GC.KeepAlive(store)` we could also use `GC.KeepAlive(store.Handle)` which would be a bit more accurate, as it's actually the `Store.Handle` instance that contains the finalizer which we need to prevent from being called.

Note that native methods that operate with the `Store` itself, like `wasmtime_store_context` (as well as other types like `Engine` etc.), are currently implemented by specifying a `Handle` instance that derives from `SafeHandle`, for which the .NET runtime uses marshalling logic to keep the `SafeHandle` instance alive during the native call. However, this requires an additional allocation for the `SafeHandle` as it's a class, and the marshalling logic might cost a bit of performance.

When optimizing for performance, it might make sense to consider replacing the `SafeHandle` with just an `IntPtr` as native method parameter, and using `GC.KeepAlive()` calls instead to keep the finalizable object alive. This would avoids the separate allocation of `SafeHandle` and would improves performance for the native call (when only using blittable types). However currently I don't see a need for this.

Thanks!

PR #207, opened 2023-01-13
Allow to define functions using "untyped callbacks"

kpreisser:
Currently, in order to define function callbacks in the `Store` or `Linker`, you have to use a `Delegate` type that specifies the signature of the function imported by WASM. This means that the function signature must be known at compile-time.
However, if you want to define callbacks at runtime without knowing the signature at compile-time, this will be difficult or even impossible.

With this PR, we allow to define functions using "untyped callbacks", which allows defining a function of any type without the need to use a specific delegate type, by using the following `UntypedCallbackDelegate`:
```cs
public delegate void UntypedCallbackDelegate(Caller caller, ReadOnlySpan<ValueBox> arguments, Span<ValueBox> results);
```

The untyped callback will receive arguments and can set results via a span of `ValueBox`. When defining the function, you will need to provide a `IReadOnlyList<ValueKind>` for the function parameters and results.

This can be thought of the callback equivalent to the `Function.Invoke()` method that allows to specify parameters as `ReadOnlySpan<ValueBox>`.

For this to work, I added new methods to `ValueBox` that allow to "unbox" the boxed value, as this was previously not possible. Note that I used the nomenclature of .NET, meaning I named the methods `AsInt32`, `AsSingle`, `AsDouble` etc., but we could also use `AsFloat32`, `AsFloat64` etc.

For example, this allows to define trapping functions for a module's imports without having to know the function signature at compile-time, similar to Wasmtime's `Linker::define_unknown_imports_as_traps` (bytecodealliance/wasmtime#4312, bytecodealliance/wasmtime#5557), like this:
```cs
// Define all imported functions using a callback that traps.
Function.UntypedCallbackDelegate trappingCallback = (caller, args, results) =>
    throw new InvalidOperationException("This function is not implemented.");

foreach (var funcImport in module.Imports.OfType<FunctionImport>())
{
    linker.DefineFunction(
        funcImport.ModuleName,
        funcImport.Name,
        trappingCallback,
        funcImport.Parameters,
        funcImport.Results);
}
```

What do you think?
Thanks!


peterhuene:
It looks like the PR needs a small rebase with the other recent merges.

kpreisser:
Thanks! I rebased the PR.

src/Function.cs
```diff
@@ -59,6 +68,51 @@ public static Function FromCallback(IStore store, Delegate callback)
...
+                Native.wasmtime_func_new(
+                    store.Context.handle,
+                    funcType,
+                    func,
+                    GCHandle.ToIntPtr(GCHandle.Alloc(func)),
+                    Finalizer,
+                    out var externFunc
+                );
+
+                return new Function(store, externFunc, parameterKinds, resultKinds);
```
peterhuene, line 113: Do we need a `GC.KeepAlive` above this?

src/Linker.cs
```diff
@@ -405,6 +405,69 @@ public void DefineFunction(string module, string name, Delegate callback)
...
+            using var funcType = Function.GetFunctionType(parameterKinds, resultKinds);
+
+            unsafe
+            {
+                Function.Native.WasmtimeFuncCallback func = (env, callerPtr, args, nargs, results, nresults) =>
+                {
+                    return Function.InvokeUntypedCallback(callback, callerPtr, args, (int)nargs, results, (int)nresults, resultKinds);
+                };
+
+                var moduleBytes = Encoding.UTF8.GetBytes(module);
```
peterhuene, line 447: Do we want to choose a stackalloc limit for these strings ala #202? Defining linker functions isn't as repeated as an operation, so I don't there's a strict need.

src/ValueBox.cs
```diff
@@ -66,6 +136,12 @@ private ValueBox ConvertTo(ValueKind convertTo)
             };
         }
 
+        private void ThrowIfNotOfCorrectKind(ValueKind expectedKind)
+        {
+            if (Kind != expectedKind)
+                throw new InvalidCastException($"Cannot convert from `{Kind}` to `{expectedKind}`");
```
peterhuene, line 142:
```suggestion
            if (Kind != expectedKind)
            {
                throw new InvalidCastException($"Cannot convert from `{Kind}` to `{expectedKind}`");
            }
```

peterhuene (approved):
Looks great 👍! Just two questions and a minor formatting fix.

src/Function.cs
kpreisser: You're right, this is needed here (I just copied the method here from the above one, and because #206 wasn't merged yet, I forgot to add the call here 😆)

src/Linker.cs
kpreisser: Agreed, I think this can then also be done for the function above with the `Delegate` parameter, so that it's consistent for all `FromCallback` overloads.

peterhuene: #202 should merge any second now if you want to rebase when it does.

peterhuene:
Ok all pending PRs are merged, so should be able to fully rebase this now. Sorry for the churn!

src/Linker.cs
kpreisser: I applied the changes from #202 to the remaining two overloads of `Linker.DefineFunction`, so it should be consistent now.

peterhuene:
Many thanks @kpreisser and @martindevans for all this work. When we merge Martin's updated to remove the alloc for `Caller`, I'll cut new releases with all these changes.

PR #209, opened 2023-01-20
Caller Unit Tests

martindevans:
Since I'm investigating changing the Caller mechanism I wanted some test coverage first. 100% coverage on `Caller`, No issues found.


peterhuene (approved):
Thanks for fixing this testing gap!

Issue #208, opened 2023-01-13
Better Caller


martindevans:
For the last couple of days I've been contemplating how to improve the `Caller` system, what I've come up with entails a large refactor of how things are done so I thought I'd lay out my thoughts and get some feedback. @peterhuene and @kpreisser I'd really value your input :)

### Context

At the moment a callback from WASM into C# can optionally take a `Caller` parameter which gives the call some context (e.g. access to `Memory` etc). For example:

```csharp
Linker.DefineFunction("env", "check_string", (Caller caller, int address, int length) =>
{
    caller.GetMemory("mem").ReadString(address, length).Should().Be("Hello World");
});
```

### The Problem

At the moment this mechanism always allocates a `Caller` for every call and will also often be used to fetch a `Memory` or `Function` which is in turn allocated. That's not great for performance (particularly in Unity, which is where I'm using wasmtime).

#### The Solution?

The obvious solution to this is to make `Caller` a `ref struct`. This also better models the intended lifetime of the `Caller` as well - you're not meant to hold it beyond that one method call.

However there is a problem with this - `Caller` implements `IStore` and in turns passes itself into `Memory` and `Function` constructors as the store reference. A `ref struct` cannot implement an interface so this doesn't work (it also can't be passed into a generic method, so we can't refactor to something like `Invoke<T> where T : IStore` either). So if we want to get a `Memory` or `Function` we'd be back needing two allocations, one for the store and one for the object itself.

To solve both problems we could introduce new memory/function types which are `ref struct`s (e.g. `RefStructMemory`) which are returned by new methods (e.g. `Caller.GetRefMemory`). Since these would be ref structs they could contain the `StoreContext` directly. Again this also models the intended lifetime better.

Obviously this is a lot of duplicated code, but I think we could reduce that by putting all of the actual work inside the `RefStructXXX` types and the current class types would become wrappers which internally create the relevant `RefStructXXX`, call into it and immediately discard it.

#### Thoughts?

I realise this is a huge amount of churn, but given that it improves the handling of lifetimes as well as performance I think it's probably worth it.

martindevans:
I just tried sketching some of this out in code and it looks like it won't work. A `ref struct` can't be used as a type parameter, so the nice type safe way we define all of our functions kills this idea I think.

kpreisser:
> A `ref struct` can't be used as a type parameter

[...]

PR #210, opened 2023-01-24
Replaced `class Caller` with `struct Caller`. Internally a `CallerCon…

martindevans:
This PR replaces the Caller class with a Caller struct, removing a potential source of a _lot_ of allocation churn. The original idea (#208) was to use a `ref struct` to properly represent the true lifetime of a caller. However, that approach was very difficult to fit into the current system and would have required extensive rewrites. Instead this system throws an `ObjectDisposedException` if you try to use things beyond their lifetime (same as the old system).

The new API is very similar to the old one and should be mostly source compatible. Two breaking changes:
1. `Caller` no longer implements `IStore`. Internally we can call `AsStore()`, but externally it's simply no longer possible. I can make that `public` if you think it's necessary, but `IStore` seems like something that shouldn't really be used externally anyway.
2. `IDisposable` is implemented explicitly. Disposal is handled internally and calling `Dispose()` is not necessary. Implementing it explicitly hides it from intellisense, but means that any using statements around a `Caller` will continue to compile.

### Future Work

At the moment getting a `Memory` from a `Caller` will allocate a `CallerStore` and `Memory`. I plan to do followup PRs to provide a workaround for this but haven't decided on the best approach yet (for example I could add `Caller.GetMemorySpan(string name, long address, int length)`).

### DIFF
--- a/src/Caller.cs
+++ b/src/Caller.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Concurrent;
 using System.Runtime.InteropServices;
 using System.Text;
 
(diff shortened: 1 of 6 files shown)


kpreisser:
Hi,
note that with the discussion from [#200](https://github.com/bytecodealliance/wasmtime-dotnet/issues/200#issuecomment-1401167856) and PR #211 (if it's accepted), it would be possible to retrieve the originating `Store` instance from the `Caller`. This also removes the `IStore` interface.

With that changes, I think the `CallerStore` and `CallerContext` classes would no longer be required, as the `Store` can just be retrieved directly from the `Caller`.
What do you think?

Thanks!

martindevans:
Storing the `Store` in the context is actually an approach I had considered for this work but I assumed the limited lifetime of objects retrieved from the `Caller` was there for a reason so I didn't want to change it.

The tricky bit of this work was tracking the lifetimes, that's why it needs the underlying `CallerContext` with an epoch number to track if it's referencing something that's disposed. However if we can directly get the `Store` there isn't any lifetime to track and the `Caller` really just becomes a thin wrapper around a `Store`. Immediate changes:
 - Remove `CallerStore` (use the main store)
 - Remove `CallerContext` (no context necessary)

Potential secondary changes:
 - `GetMemory`/`GetFunction` could have a cache in/alongside the `Store`, that way getting things in multiple calls only allocates the first time.

**Tl;dr: We should decide on #211 before reviewing this PR.**

peterhuene:
Thanks, very much appreciated!

peterhuene:
Closing this for the impending re-implementation (thanks for your patience on this @martindevans!).

PR #211, opened 2023-01-24
Store the `Store` instance itself in the context data

### LINKED ISSUE #200: `Function` instance passed to a callback (as `funcref`) cannot be called outside of the callback

kpreisser:
Store the `Store` instance itself in the context data using a weak `GCHandle`, so that it can be retrieved later from the `Caller.Store` property.

By using a **weak** `GCHandle`, we avoid the issue that the circular reference between the `Store` and the `GCHandle` would prevent the `Store` from being GCed if it's no longer reachable from user code.

Additionally (as suggested in https://github.com/bytecodealliance/wasmtime-dotnet/issues/200#issuecomment-1401167856), remove the `IStore` interface and change the corresponding method/constructor parameters to accept a `Store` instead.

For example, this allows to create a `funcref` value as `Function` using the associated `Store` instead of the `Caller`, so that it can be stored and then called outside of the callback.

**Note:** Users that previously passed a `Caller` instance to a method accepting an `IStore` will now need to use `caller.Store`.

This is a remake of the earlier PR, #201.

Fixes #200

What do you think?
Thanks!


src/Store.cs
```diff
@@ -28,32 +28,20 @@ internal void AddFuel(ulong fuel)
...
-            }
+                var data = Native.wasmtime_context_get_data(handle);
 
-            Native.wasmtime_context_set_data(handle, newPtr);
+                // Since this is a weak handle, it could be `null` if the target object (`Store`)
+                // was already collected. However, this would be an error in wasmtime-dotnet
+                // itself because the `Store` must be kept alive when this is called, and
+                // therefore this shouldn't happen.
+                var targetStore = (Store?)GCHandle.FromIntPtr(data).Target ??
+                    throw new ObjectDisposedException(nameof(Wasmtime.Store));
```
kpreisser, line 42: Thinking more about this, I think the check for `null` here (and throwing in that case) probably doesn't make much sense because if the `Store` would actually have been GCed, its `Handle` might also already have been deleted, which in turn would have freed the `GCHandle`, and thus it would already be invalid (leading to UB) to try to retrieve the `GCHandle` from the `IntPtr` value in such a case.

kpreisser: I removed the null check here.

peterhuene (approved):
This looks good, I'm happy with were we landed with this.

@martindevans would you also mind reviewing to see how this impacts #210?

martindevans:
It looks good to me 👍 

If you go ahead with merging this then I'd suggest simply closing #210 for now - this changes the landscape enough that it'll be eaisest for me to just reimplement everything on top of it.

peterhuene:
@martindevans do you think you'd be able to replace #210 today? I am hoping to clear out the PR queue and cut some releases that are long overdue.

martindevans:
I'll give it a go. I think this should significantly simplify things, so hopefully I can get it done within an hour or two. If it looks like it's going to be more complex and I can't get it done within that window I'll comment on #210.

peterhuene:
@martindevans Thanks, very much appreciated!

martindevans:
Ah I think this may have a bug, apologies for not catching it before the merge. It looks like it fails one of my new Caller tests (introduced in #209).

This test triggers an access violation exception:

```csharp
[Fact]
public void ItCannotBeAccessedAfterDisposal()
{
    Caller stash = default;

    Linker.DefineFunction("env", "callback", (Caller c) =>
    {
        stash = c;
    });

    var instance = Linker.Instantiate(Store, Fixture.Module);
    var callback = instance.GetFunction("call_callback")!;

    callback.Invoke();

    var act = () => stash.GetMemory("memory");
    act.Should().Throw<ObjectDisposedException>();
}
```

Dispose is called internally but of course I've already stashed away a _copy_ before that happens, so that `Caller` now has invalid state.

If no one else has any other solutions I can fix this by adding back in my epoch based mechanism?

kpreisser:
Hi @martindevans,
I'm not sure if I understand the issue; do you mean that when the `Caller` is changed to a `struct` instead of a `class`, the access violation may happen?

Currently, in `main` with the latest commit (43a5b51), `Caller` is still a class, and therefore when the callback returns, it should set its `Handle` to `IntPtr.Zero`, so that when later calling `stash.GetMemory(...)`, it should throw an `ObjectDisposedException`, which seems also to happen for me (as the tests pass).

Thanks!

martindevans:
> I'm not sure if I understand the issue; do you mean that when the Caller is changed to a struct instead of a class, the access violation may happen?

Derp, that's why CI didn't catch it, because it's my own bug. Sorry 🤦‍♂️

PR #213, opened 2023-01-25
Remove slow reflection-based fallback for function callbacks.

peterhuene:
This PR removes the slow reflection-based fallback that is performed when a callback delegate is provided that does not conform to the (up to) 12 parameter and (up to) 4 results overloads of `Function.FromCallback` and `Linker.DefineFunction`.

As a result, functions that take more than 12 parameters or return more than 4 results must use an untyped callback where the parameters and results are passed as spans of `ValueBox`.

The callbacks will no longer be invoked via reflection under any circumstances.


kpreisser (commented):
I think this looks good!

One thing I noticed is, `GetFunctionType` now no longer takes an `allowTuple` parameter which specifies whether the `Func<...>` is allowed to use a `ValueTuple<...>` as return parameter.

This means when you use the "wrong" overload to define a function returning multiple values, `Function.GetFunctionType()` will succeed but `ValueRaw.Converter<TResult>()` will fail. This results in an `InvalidOperationException` with a message that might be a bit unintuitive, for example:

```cs

linker.DefineFunction("", "func1", () => (1, 2)); // Succeeds
linker.DefineFunction<int, int>("", "func2", () => (1, 2)); // Succeeds

// Throws "System.InvalidOperationException: 'Cannot convert type 'ValueTuple`2' into a WASM parameter type'"
linker.DefineFunction<(int, int)>("", "func3", () => (1, 2)); 

// For comparison (here, Function.GetFunctionTypes() fails):
// Throws "Wasmtime.WasmtimeException: 'Unable to create a callback with a return type of type 'System.ValueTuple`2[System.Int32,System.Int32]'.'"
linker.DefineFunction<((int, int), int)>("", "func3", () => ((1, 2), 2));

```

When keeping the `allowTuple` parameter here, we could do something like:
```cs
if (returnsTuple && !allowTuple)
{
    throw new InvalidOperationException($"Returning a ValueTuple is not allowed for this callback; use an overload that implicitly returns ValueTuple.");
}
```

(or a another/better message). This will ensure that the parameter/result types are validated by `Function.GetFunctionType()` before they reach `IValueRaw.Converter()`.

Additionally, in that case I think the `maxResults` parameter might not be needed any more.

What do you think?
Thanks!

peterhuene:
@kpreisser Thanks for the feedback! You're right, the change to guide users towards an untyped callback in cases where there were too many results inadvertently let the invalid types slip through to value conversion causing the more confusing error message that should have been caught at function definition time.

Your suggestion makes the most sense and provides a really great error message. I'll push a fix.

peterhuene:
@kpreisser I've pushed up the suggested change. Would you mind reviewing if you have the time?

src/Function.FromCallback.tt
```diff
@@ -54,52 +53,40 @@ foreach (var (hasCaller, resultCount, parameterCount, methodGenerics, delegateTy
...
-                Native.wasmtime_func_new_unchecked(
-                    store.Context.handle,
-                    funcType,
-                    func,
-                    GCHandle.ToIntPtr(GCHandle.Alloc(func)),
-                    Finalizer,
-                    out var externFunc
-                );
+                var funcType = CreateFunctionType(parameterKinds, resultKinds);
+                var externFunc = new ExternFunc();
```
kpreisser, line 68: Nit: If the `out` modifier for the `Externfunc` parameter in `wasmtime_func_new_unchecked` is correct (which means the function will always write the value), it should suffice to just declare the variable:
```suggestion
                ExternFunc externFunc;
```

Otherwise (if the called function does not always write the value) it should be declared as `ref` parameter instead (but I assume the `out` declaration is correct here).

kpreisser (approved):
Thank you! Looks good to me.

src/Function.FromCallback.tt
peterhuene: Fixed it here and in one other place.

PR #214, opened 2023-01-25
Ref struct caller

martindevans:
~~**This builds on #213, merge that first**~~

Third time lucky?

Replaced `Caller` with a `ref struct`. With the reflection path stripped out and the removal of `IStore` this was a fairly simple change.

A few notable changes:
 - Can't leak `Caller`, any breakage here is probably uncovering a bug.
 - `Caller` is no longer disposable (we could add an empty `Dispose()` method and mark it `[Obsolete]`?).

Also this test:

```
[Fact]
public void ItCannotBeConstructedFromNullPointer()
{
    var act = () => new Caller(IntPtr.Zero);
    act.Should().Throw<InvalidOperationException>();
}
```

~~Did not compile because `act` is now a `CallerAction` instead of an `Action`, which is surprising change of `var` inference. Fixed by changing `var` to `Action`.~~ Never mind, see comment by kpreisser below.

## Benchmark

Unlike my previous epoch based approach in #212 (which was 90ns slower) this benchmarks at exactly the same speed as the `class Caller` approach.

### DIFF
--- a/src/Caller.cs
+++ b/src/Caller.cs
@@ -7,7 +7,7 @@ namespace Wasmtime
     /// <summary>
     /// Represents caller information for a function.
     /// </summary>
-    public class Caller : IDisposable
+    public readonly ref struct Caller
     {
         internal Caller(IntPtr handle)
         {
@@ -17,7 +17,8 @@ namespace Wasmtime
             }
 
             this.handle = handle;
-            store = Context.Store;
+            this.context = new StoreContext(Native.wasmtime_caller_context(handle));
+            this.store = context.Store;
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


martindevans:
I thought I'd let the time zone differential work in my favour today ;)

kpreisser:
> Did not compile because `act` is now a `CallerAction` instead of an `Action`

I think this is unrelated to the new `CallerAction<...>` and `CallerFunc<...>` types; it's actually an anonymous delegate type that the C# compiler creates when it can't match an existing delegate type (`Action`/`Func`):
grafik

This is because it would normally infer a `Func<Caller>` (as the expression `new Caller(...)` can also be used as `return` expression), which can't be matched now due to the `ref struct Caller`, so it generates a new delegate type like `delegate Caller X()`.

peterhuene:
This looks good; I guess the explicitly defined delegates weren't as bad as I thought they'd be.

Let's merge #213, rebase this PR, and then I'll review it for inclusion in the next release.

peterhuene:
We should be able to rebase this now.

martindevans:
Rebased onto main.

peterhuene (approved):
Awesome! Thanks for this and I'm glad we can now leave the escaping of `Caller` from the current frame up to the C# compiler 👍.

peterhuene:
Okay, with everything merged, I'm going to cut releases and call out any potential API incompatibilities.

PR #221, opened 2023-02-06
Fix the .NET API for recent Wasmtime C API changes.

peterhuene:
This PR fixes the definition of `wasmtime_linker_define` to match what is currently expected upstream.

As a result, items passed to `Linker.Define` must be associated with a store.

The internal `IExternal` interface was updated to expose the store associated with the item for `LinkerDefine` to use.

Also update the target framework to 7.0 (still `netstandard2.1` as well).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: '6.0.x'
+        dotnet-version: '7.0.x'
     #  workaround for actions/setup-dotnet#155
     - name: Clear package cache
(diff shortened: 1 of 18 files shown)


peterhuene:
I'm going to see if I can expose the root cause for the flaky tarball downloads from GitHub that seems to only impact the Windows CI machines given it's still happening.

peterhuene:
> error MSB3923: Failed to download file "https://github.com/bytecodealliance/wasmtime/releases/download/dev/wasmtime-dev-aarch64-linux-c-api.tar.xz".  Authentication failed, see inner exception.

Sigh. The inner exception seems to be completely eaten by the task and not logged in anyway at even a diagnostic logging level.

peterhuene:
I'm going to merge this for now and we'll figure out the transient download issues on the Windows CI later (it'll require implementing our own `DownloadFile` task to get the full exception message).

PR #222, opened 2023-02-07
Add a Retries="10" attribute to the `DownloadFile` task

kpreisser:
Add a `Retries="10"` attribute to the `DownloadFile` task, to check whether it can avoid the failed download issue in the Windows CI runner.

**Edit:** Unfortunately it seems `IsRetriable` returns false, so the task doesn't attempt to retry the download.


peterhuene:
Yeah, unfortunately the underlying exception isn't "retryable".

The good news is that there is a [fix to the `DownloadFile` task](https://github.com/dotnet/msbuild/pull/8440) that might expose what the underlying issue is with our flaky Windows CI (my money is on a TLS issue with GitHub).

PR #224, opened 2023-02-15
Converted `Memory` into a `readonly struct`

martindevans:
This is a demonstration of what's involved with converting `Memory` into a `readonly struct`, as discussed in (https://github.com/bytecodealliance/wasmtime-dotnet/pull/219#pullrequestreview-1296613008). As you can see this was almost no work. The bulk of the changes are fixing the tests.

My only concern with this is that it is now possible to call things on a `default(Memory)`, which of course would have simply triggered a null reference exception before. I've checked through and I think it's safe in this case because everything uses `store.Context` which will trigger a null reference exception.

Thoughts?

### DIFF
--- a/examples/memory/Program.cs
+++ b/examples/memory/Program.cs
@@ -11,7 +11,7 @@ linker.Define(
     "log",
     Function.FromCallback(store, (Caller caller, int address, int length) =>
     {
-        var message = caller.GetMemory("mem").ReadString(address, length);
+        var message = caller.GetMemory("mem")?.ReadString(address, length);
         Console.WriteLine($"Message from WebAssembly: {message}");
     }
 ));
--- a/src/Memory.cs
+++ b/src/Memory.cs
@@ -8,7 +8,7 @@ namespace Wasmtime
     /// <summary>
     /// Represents a WebAssembly memory.
     /// </summary>
-    public class Memory : IExternal
+    public readonly struct Memory : IExternal
     {
         /// <summary>
         /// Creates a new WebAssembly memory.
@@ -553,6 +553,7 @@ namespace Wasmtime
         {
             this.memory = memory;
             this.store = store;
+            this.Maximum = null;
 
             var typeHandle = Native.wasmtime_memory_type(store.Context.handle, this.memory);
             try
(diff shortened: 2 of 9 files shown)


kpreisser (commented):
I like it! (Maybe this could later also be done for other `Extern` types.)

One minor thing to note is that the `Memory` struct size (`Unsafe.SizeOf<Memory>()`) is currently `56` due to the fields `Minimum`, `Maximum` and `Is64Bit`, which can also be retrieved from the `ExternMemory`. Removing the fields would reduce the struct size to `24`, in which case the properties would need to retrieve the values from the `ExternMemory` on every call. But I'm not sure how expensive that is, so it's probably OK to keep it in the struct.

Also note that because `Memory` implements an interface (`IExternal`), this will cause boxing to occur where this is used, for example with `Linker.Define(string module, string name, object item)`, so an additional allocation will now occur there.
If desired, I think this could be avoided by having an internal `Linker.Define(string, string, Store store, Extern extern)`, and then overloads for the struct types like `Linker.Define(string, string, Memory)` that calls it.

Thanks!

martindevans:
> this will cause boxing to occur where this is used, for example with `Linker.Define(string module, string name, object item)`

If it's just an internal interface I would change it to `Linker.Define<T>(string module, string name, T item) where T : IExtern`. Which would also have the benefit that you can't pass non-externs into it (which is currently checked at runtime). That won't work for the public API though since `IExtern` is internal. For the public API can we just add overloads for all the individual extern things, there are only 4 (Memory, Table, Global, Module) as far as I know?

peterhuene:
> For the public API can we just add overloads for all the individual extern things, there are only 4 (Memory, Table, Global, Module) as far as I know?

That sounds reasonable to me!

martindevans:
MacOS CI failure is a bit cryptic, I'm hoping it's just another transient error.

kpreisser:
Hi,
it seems you intended to rebase the PR on current `main` branch (resulting in 7e2d02a8ae896e63271d8e9c89b1acfa7efcc841), but then merged it with the previous commit of this PR (617be6fd5fb7c407d6fac36b6e03ed60ff12a533), which is why the Github UI is now showing changes that are already part of `main`.

Can you reset the branch to commit 7e2d02a8ae896e63271d8e9c89b1acfa7efcc841 and force-push it (as I think that was the intended commit)?
Thanks!

martindevans:
Unfortunately I don't have access to a Mac to investigate the test issues. There's no detail in the log on what the problem might be either :/

kpreisser:
I'm also not sure why it's crashing, but maybe it is caused by/related to the `MemoryAccessTests` that allocate a 4 GB Memory (`ItCanAccessMemoryWith65536Pages` + `ItThrowsForOutOfBoundsAccess`).
Can you try to add the same attribue as for the `Memory64AccessTests` to skip them, to check if that avoids the crash?  Though I'm not sure why this would happen on this PR only.

https://github.com/bytecodealliance/wasmtime-dotnet/blob/9562c78652208c49e55c6618d4ff64103dfec9e9/tests/Memory64AccessTests.cs#L28

Thanks!

martindevans:
I've added skips to three tests:
- ItCanAccessMemoryWith65536Pages
- ItThrowsForOutOfBoundsAccess
- AccessDefaultThrows

martindevans:
Superseded by #235, which allows `Memory` to remain as a class.

PR #228, opened 2023-02-28
Extern define

martindevans:
- Added public `Define` methods for all externs.
 - Removed method which accepted an `object` parameter and dynamically checked that it was an extern.

As discussed in https://github.com/bytecodealliance/wasmtime-dotnet/pull/224#issuecomment-1433591490


peterhuene (approved):
Looks good, thanks!

PR #233, opened 2023-03-16
`Caller.TryCallFunction`

martindevans:
Added `TryCallFunction` methods to `Caller`. This allows functions to be called without allocating a `Function` object. An alternative solution would be to convert `Function` into a struct, but this is a much larger change!

The main shortcoming with this is that `ValueBox` cannot be stack allocated. So this API is really just passing the buck on allocating anything - the caller _can_ avoid allocations (rent an array) but it's not very ergonomic. See the unit tests for example.

Another implementation could have generic arguments for all possible call/return combinations (and then use `ValueBox.Converter<T>().Box(value)`). This would be a fairly nice API to use, but would require generating yet more huge blobs of code.

I'm open to other suggestions on a better way to pass arguments and results.

### DIFF
--- a/src/Caller.cs
+++ b/src/Caller.cs
@@ -102,6 +102,30 @@ namespace Wasmtime
             }
         }
 
+        private ExternFunc? TryGetFunc(string name)
+        {
+            unsafe
+            {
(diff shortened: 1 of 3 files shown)


martindevans:
Superseded by #235

PR #235, opened 2023-03-24
Caller Function Cache

martindevans:
Added caching to the `Store` for `Function` and `Memory` objects. The `Caller` uses this to retrieve `Function`/`Memory` objects, this saves allocating lots of objects in caller contexts.

This is an alternative to #233 and #224. I prefer this approach to #233 since it gives access to the entire `Function` API, instead of a tiny subset exposed through the `Caller`. I prefer it to #224 since the `Memory` object remains a class instead of becoming a struct, which is more idiomatic.

### DIFF
--- a/src/Instance.cs
+++ b/src/Instance.cs
@@ -524,7 +524,7 @@ namespace Wasmtime
 
             GC.KeepAlive(_store);
 
-            return new Function(_store, ext.of.func);
+            return _store.GetCachedExtern(ext.of.func);
         }
 
         /// <summary>
--- a/src/Instance.cs
+++ b/src/Instance.cs
@@ -557,9 +557,7 @@ namespace Wasmtime
                 return null;
             }
 
-            GC.KeepAlive(_store);
-
-            return new Memory(_store, ext.of.memory);
+            return _store.GetCachedExtern(ext.of.memory);
         }
 
         /// <summary>
--- a/src/Instance.cs
+++ b/src/Instance.cs
@@ -557,6 +557,8 @@ namespace Wasmtime
                 return null;
             }
 
+            GC.KeepAlive(_store);
+
             return _store.GetCachedExtern(ext.of.memory);
         }
 
--- a/src/Linker.cs
+++ b/src/Linker.cs
@@ -289,7 +289,7 @@ namespace Wasmtime
                         throw WasmtimeException.FromOwnedError(error);
                     }
 
-                    return new Function(store, func);
+                    return store.GetCachedExtern(func);
                 }
             }
         }
@@ -316,7 +316,7 @@ namespace Wasmtime
 
             GC.KeepAlive(store);
 
-            return new Function(store, ext.of.func);
+            return store.GetCachedExtern(ext.of.func);
         }
 
         /// <summary>
--- a/src/Linker.cs
+++ b/src/Linker.cs
@@ -364,6 +364,8 @@ namespace Wasmtime
                 return null;
             }
 
+            GC.KeepAlive(store);
+
             return store.GetCachedExtern(ext.of.memory);
         }
 
--- a/src/Linker.cs
+++ b/src/Linker.cs
@@ -364,9 +364,7 @@ namespace Wasmtime
                 return null;
             }
 
-            GC.KeepAlive(store);
-
-            return new Memory(store, ext.of.memory);
+            return store.GetCachedExtern(ext.of.memory);
         }
 
         /// <summary>
--- a/src/Store.cs
+++ b/src/Store.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using Microsoft.Win32.SafeHandles;
 
@@ -164,9 +163,6 @@ namespace Wasmtime
         /// </summary>
         public void GC()
         {
-            _funcCache.Clear();
-            _memCache.Clear();
-
             Context.GC();
             System.GC.KeepAlive(this);
         }
--- a/src/Store.cs
+++ b/src/Store.cs
@@ -306,32 +306,32 @@ namespace Wasmtime
 
         private static readonly Native.Finalizer Finalizer = (p) => GCHandle.FromIntPtr(p).Free();
 
-        private readonly ConcurrentDictionary<(ulong store, nuint index), Function> _funcCache = new();
+        private readonly ConcurrentDictionary<(ExternKind kind, ulong store, nuint index), object> _externCache = new();
+
         internal Function GetCachedExtern(ExternFunc @extern)
         {
-            var key = (@extern.store, @extern.index);
+            var key = (ExternKind.Func, @extern.store, @extern.index);
 
-            if (!_funcCache.TryGetValue(key, out var func))
+            if (!_externCache.TryGetValue(key, out var func))
             {
                 func = new Function(this, @extern);
-                func = _funcCache.GetOrAdd(key, func);
+                func = _externCache.GetOrAdd(key, func);
             }
 
-            return func;
+            return (Function)func;
         }
 
-        private readonly ConcurrentDictionary<(ulong store, nuint index), Memory> _memCache = new();
         internal Memory GetCachedExtern(ExternMemory @extern)
         {
-            var key = (@extern.store, @extern.index);
+            var key = (ExternKind.Memory, @extern.store, @extern.index);
 
-            if (!_memCache.TryGetValue(key, out var mem))
+            if (!_externCache.TryGetValue(key, out var mem))
             {
                 mem = new Memory(this, @extern);
-                mem = _memCache.GetOrAdd(key, mem);
+                mem = _externCache.GetOrAdd(key, mem);
             }
 
-            return mem;
+            return (Memory)mem;
         }
     }
 }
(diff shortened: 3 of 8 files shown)


martindevans:
@kpreisser I've updated this PR with some extra work, basically adding the same caching for `Memory` as well as `Function`. That means it supersedes #224 as well (and we can conveniently ignore the MacOS CI weirdness).

I'd appreciate your re-review :)

src/Linker.cs
```diff
@@ -364,9 +364,7 @@ public Function GetDefaultFunction(Store store, string name)
                 return null;
             }
 
-            GC.KeepAlive(store);
-
-            return new Memory(store, ext.of.memory);
+            return store.GetCachedExtern(ext.of.memory);
```
kpreisser, line 367: I would prefer to keep the `GC.KeepAlive(store)` call here for consistency, as the above code access the `StoreContext`.
Otherwise, when making changes at this location later you would have to reason about whether the `store` is still accessed, or the compiler/GC could optimize it away and finalize the `Store` (like a code debt). What do you think?

kpreisser (commented):
I added a few notes, but generally this looks good to me, thanks!

However, we will probably need to wait for @peterhuene, to verify e.g. that the usage of only the `index` field of the `ExternXyz` is correct/sufficient to match the object from the dictionary (I would think so, as the dictionary will only store objects that belong to the current `Store`), or if it would technically be required to compare the whole `ExternXyz` (`store` + `index`).

src/Linker.cs
martindevans: Added this back in 👍

src/Store.cs
```diff
@@ -162,6 +164,9 @@ public Store(Engine engine, object? data)
         /// </summary>
         public void GC()
         {
+            _funcCache.Clear();
```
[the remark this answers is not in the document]
martindevans: Yeah I wasn't too sure about this either. I wanted to have _some_ way to preventing infinite growth of the `Store` and `GC()` seemed like the best existing option, but it's not exactly a perfect fit.

You're correct about things living forever in the `Store`, to quote the Wasmtime docs:

> A [Store](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html) is intended to be a short-lived object in a program. No form of GC is implemented at this time so once an instance is created within a [Store](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html) it will not be deallocated until the [Store](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html) itself is dropped. This makes [Store](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html) unsuitable for creating an unbounded number of instances in it because [Store](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html) will never release this memory. It’s recommended to have a [Store](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html) correspond roughly to the lifetime of a “main instance” that an embedding is interested in executing.

Given this I've removed the cache clearing for now. We can look into more cache control in the future if necessary.

martindevans:
I've addressed those three comments.

> However, we will probably need to wait for @peterhuene, to verify e.g. that the usage of only the index field of the ExternXyz is correct/sufficient to match the object from the dictionary (I would think so, as the dictionary will only store objects that belong to the current Store), or if it would technically be required to compare the whole ExternXyz (store + index).

I would like to add a sanity check that `extern.store == this_store`, but I'm not sure exactly what I should compare with. The docs simply say it's an "internal identifier" and as far as I can see there's no way to retrieve that identifier for a given `wasmtime_store*`.

peterhuene:
So I don't think you'll be able to assert on store ids, but could we assert that the item (memory/function) being return from the cache has the same store object reference as `this`?

peterhuene (commented):
LGTM.

I'll merge after @kpreisser completes their review.

martindevans:
I don't think checking the `Store` field of the `Memory`/`Function` will help much, since those objects are all constructed inside the `Store` (referencing `this`) in the first place.

If it's not possible to check then I think we're probably good enough as we are. Any objects constructed in the cache will use the store context internally and will presumably detect errors for us. For example `Memory` does this:

```Native.wasmtime_memory_type(store.Context.handle, this.memory);```

Which I assume will internally sanity check if the store context and the memory are mismatched?

**Comment by kpreisser** [comment]
👍 

Additionally, to be on the safe side regarding the `uint64_t store_id` (`ulong store`) value of externs that Wasmtime uses, we could use both fields from the `Extern...` as key (`ValueTuple`) for the dictionary, so that we don't need to assume that Wasmtime always uses the same `ulong store` value for `Extern...` objects for the current store, but instead we simply use the same values to identify the extern as Wasmtime does. E.g.
```cs
private readonly ConcurrentDictionary<(ulong store, nuint index), Function> _funcCache = new();
internal Function GetCachedExtern(ExternFunc @extern)
{
    var key = (@extern.store, @extern.index);
    if (!_funcCache.TryGetValue(key, out var func))
    {
        func = new Function(this, @extern);
        func = _funcCache.GetOrAdd(key, func);
    }

    return func;
}
```

This should work as `ValueTuple<...>` provides an implementation of `Equals` and `GetHashCode` that combines the members of the tuple. 

What do you think?
Thanks!

martindevans:
I've changed the key to a tuple of store/index as @kpreisser suggested.

kpreisser (approved):
Would you mind adding tests that e.g. verify that two consecutive calls of `Caller.GetFunction`, `Caller.GetMemory` with the same name actually return the same `Function`/`Memory` instance (so the cache works)?

Apart from that and the comment, this looks good to me, thanks!

A possible optimization idea for the future (especially when a cache is also added for other externs like `Table`, would be to check whether the multiple `ConcurrentDictionary`s can be combined into a single one, where the key additionally consists of `ExternKind`, because AFAIK creating a `ConcurrentDictionary` involes [a number of allocations](https://github.com/dotnet/runtime/blob/d0913fce8975849f90fd232698d16c3f95d89d6d/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L180-L223), which could then be saved.

src/Instance.cs
```diff
@@ -557,9 +557,7 @@ public Instance(Store store, Module module, params object[] imports)
                 return null;
             }
 
-            GC.KeepAlive(_store);
```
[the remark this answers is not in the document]
martindevans: I've added that back in.

martindevans:
> Would you mind adding tests that...

I've added tests for:
 - Getting memory multiple times always returns the same object.
 - Getting a non-existant memory returns null, even after getting something else from the cache.
 - Getting a function multiple time always returns the same object.
 - Getting two _different_ functions does _not_ return the same object.

> A possible optimization idea ... key of ExternKind

I did look into this but it's made a bit difficult by the way `Extern` is built at the moment - the `ExternUnion` doesn't know which variant it represents so equality/GetHashCode can't be implemented on it.

I'll probably come back to look at this again soon. It looks like it can be fixed by merging the fields of `ExternUnion` into `Extern` so that all the necessary information is in one place. This would create lots of small changes across the rest of the codebase, so I didn't want to do that in this PR.

kpreisser:
Thanks!

> I did look into this but it's made a bit difficult by the way `Extern` is built at the moment - the `ExternUnion` doesn't know which variant it represents so equality/GetHashCode can't be implemented on it.

Actually, what I meant was to use something like 
```cs
ConcurrentDictionary<(ExternKind kind, ulong store, nuint index), IExternal> _externCache
```
 i.e. add the `ExternKind` to the key, which I think should work since all externs have the same fields (`ulong store` and `nuint index`), so that only a single `ConcurrentDictionary` would be needed. The functions could then look like this:

```cs
internal Function GetCachedExtern(ExternFunc @extern)
{
    var key = (ExternKind.Func, @extern.store, @extern.index);

    if (!_externCache.TryGetValue(key, out var func))
    {
        func = new Function(this, @extern);
        func = _externCache.GetOrAdd(key, func);
    }

    return (Function)func;
}

internal Memory GetCachedExtern(ExternMemory @extern)
{
    var key = (ExternKind.Memory, @extern.store, @extern.index);

    if (!_externCache.TryGetValue(key, out var mem))
    {
        mem = new Memory(this, @extern);
        mem = _externCache.GetOrAdd(key, mem);
    }

    return (Memory)mem;
}
```

However, a downside is the additional cast e.g. to `Function` or `Memory`, since the dictionary's value type parameter is `IExtern`.
What do you think?

tests/CallerTests.cs
```diff
@@ -47,6 +47,30 @@ public void ItCanGetMemoryFromCaller()
         callback.Invoke();
     }
 
+    [Fact]
+    public void ItCanGetCachedMemoryFromCaller()
+    {
+        var memories = new HashSet<Memory>();
```
kpreisser, line 53: Nit: To guarantee that the `HashSet` checks for object reference equality only, even if `Memory` would override `Equals`/`GetHashCode` in the future, you can pass `ReferenceEqualityComparer.Instance` in the constructor.

martindevans: Fixed

martindevans:
> Actually, what I meant was...

That's much simpler than what I had in mind, good idea. I've added that.

peterhuene (approved):
Looks good, thanks again for contributing this!

PR #240, opened 2023-04-04
Added a cache of the Global.Accessor

martindevans:
Added a cache for the Global type safe accessor, similar to the wrapped function cache introduced #236.


peterhuene (approved):

peterhuene:
Thanks for adding this!

PR #243, opened 2023-04-24
Implemented `Store.SetLimits`

martindevans:
Implemented `Store.SetLimits` using the new `wasmtime_store_limiter` function.

### DIFF
--- a/src/Store.cs
+++ b/src/Store.cs
@@ -196,7 +196,9 @@ namespace Wasmtime
 
             long tableElements64 = -1;
             if (tableElements.HasValue)
+            {
                 tableElements64 = tableElements.Value;
+            }
 
             Native.wasmtime_store_limiter(NativeHandle, memorySize ?? -1, tableElements64, instances ?? -1, tables ?? -1, memories ?? -1);
         }
--- a/src/Store.cs
+++ b/src/Store.cs
@@ -159,6 +159,25 @@ namespace Wasmtime
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


kpreisser:
Thanks for creating this PR! I also thought about implementing this, but didn't find time to do so.

> Opinions?

I would favor option 2 (use nullable values), at least for the parameters that can specify an unlimited value according to the documentation (`memory_size` and `table_elements`). This would match e.g. the `Memory.Maximum` API which is also a nullable `long?`, where `null` means the memory doesn't have a maximum.

Regarding accepting both negative values and `null`, we could throw an `ArgumentOutOfRangeException` if a negative value is specified, so that `null` is the only way to specify that the default value should be used.

What do you think?
Thanks!

src/Store.cs
```diff
@@ -158,6 +158,49 @@ public Store(Engine engine, object? data)
...
+                throw new ArgumentOutOfRangeException(nameof(tables));
+            }
+
+            if (memories.HasValue && memories.Value < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(memories));
+            }
+
+            long tableElements64 = -1;
+            if (tableElements.HasValue)
```
peterhuene, line 198: Style nit: braces 😅

peterhuene (approved):
Looks great, just the incredibly minor nit!

src/Store.cs
martindevans: Fixed 👍

peterhuene:
Thanks for implementing this!

Issue #39, opened 2020-10-01
wasmtime NetStandard 2.0


onehundredfeet:
Hi,

In order to use this inside Unity, I've 'ported' the system to NetStandard2.0. It was a fairly quick port, but didn't modify too much of the code. I would be much more comfortable if there was an official port. Or #if 's in the code to handle it.

I'm happy to put my code somewhere if you want to see it, but I'm sure it's inferior to what the author of the module would do.

Cheers,
Ryan.

peterhuene:
Hi @onehundredfeet.  Sorry for the delayed response as I've had most of October off.

The biggest hurdles with supporting netstandard 2.0 is the lack of `Span`, `BinaryPrimatives`, and nullable reference types (i.e. C# 8.0).

There are some unsupported hacks one can do to enable C# 8.0 with a `netstandard2.0` TFM, but I'd rather not do that (or lots of `#if`s in the code if we can avoid either).

However, I don't want this to be a barrier to adoption, but it does seem that the Unity folks are busy updating to a more recent version of Mono which should bring netstandard 2.1 support (maybe even `net5.0` if we're lucky?).

Perhaps the best course of action is to wait and see if Unity will release a new version soon rather than hack this library to work with a 3 year old TFM.

onehundredfeet:
No worries. 

Most of what is needed, span etc, has been made available with system.memory. I’ve successfully got wasmtime up and running in unity with minimal changes. I’ve checked it into my branch if you’re ever curious.

zalo:
> I’ve checked it into my branch if you’re ever curious.

Hi @onehundredfeet , sorry to be a bother, but have your version of wasmtime uploaded anywhere?

It doesn't seem like the requisite types are present in 2020.2 (despite supposedly being C#8...)

jazzay:
@onehundredfeet are you able to share your Unity specific branch?

alexhiggins732:
PR #94 upgrades to .Net 6? Does that address this issue so it can be closed?

PR #246, opened 2023-04-30
.NET Standard 2.0 support

IS4Code:
Updated to support .NET Standard 2.0 via multi-targeting. This brings the following changes (most of which are conditionally-compiled):

* Added references to `System.Memory` and `IndexRange` (only for `netstandard2.0`).
* `UnmanagedType.LPUTF8Str` is referenced as `(UnmanagedType)48`.
* `Marshal.PtrToStringUTF8` and various other methods are reimplemented using various workarounds, found in the `Extensions` class.
* Detection of tuple types checks for types starting with <code>System.ValueTuple\`</code> instead of implementing `ITuple` (which is perhaps even better, since `System.Tuple` also implements `ITuple` but cannot be used this way).
* Static methods in `IReturnTypeFactory` moved to new non-interface `ReturnTypeFactory`.
* Unconditional workarounds for `GetSubArray` (should be faster anyway).

### DIFF
--- a/src/Extensions.cs
+++ b/src/Extensions.cs
@@ -1,0 +1,116 @@
+﻿using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Wasmtime
+{
+    internal static class Extensions
+    {
+        public const UnmanagedType LPUTF8Str =
+#if NETSTANDARD2_0
+            (UnmanagedType)48;
+#else
+            UnmanagedType.LPUTF8Str;
+#endif
+
+#if NETSTANDARD2_0
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe string GetString(this Encoding encoding, Span<byte> bytes)
+        {
+            fixed (byte* bytesPtr = bytes)
+            {
+                return encoding.GetString(bytesPtr, bytes.Length);
+            }
+        }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe string GetString(this Encoding encoding, ReadOnlySpan<byte> bytes)
+        {
+            fixed (byte* bytesPtr = bytes)
+            {
+                return encoding.GetString(bytesPtr, bytes.Length);
+            }
+        }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe int GetBytes(this Encoding encoding, Span<char> chars, Span<byte> bytes)
+        {
+            fixed (char* charsPtr = chars)
+            fixed (byte* bytesPtr = bytes)
+            {
+                return encoding.GetBytes(charsPtr, chars.Length, bytesPtr, bytes.Length);
+            }
+        }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe int GetBytes(this Encoding encoding, ReadOnlySpan<char> chars, Span<byte> bytes)
+        {
+            fixed (char* charsPtr = chars)
+            fixed (byte* bytesPtr = bytes)
+            {
+                return encoding.GetBytes(charsPtr, chars.Length, bytesPtr, bytes.Length);
+            }
+        }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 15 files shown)


peterhuene:
Hi @IS4Code. Thanks for the PR! Rebasing this PR on the latest in `main` should fix the CI issue.

We'll review soon.

src/V128.cs
```diff
@@ -92,7 +92,14 @@ public Span<byte> AsSpan()
         {
             unsafe
             {
+#if NETSTANDARD2_0
+                fixed (byte* bytesPtr = bytes)
+                {
+                    return new Span<byte>(bytesPtr, 16);
+                }
```
kpreisser, line 99: Note that on platforms without built-in `Span` support like .NET Framework, this will only be valid if the `V128` is on the stack, because the GC won't be able to track a reference to an object field.

For example, with a code like the following:
```cs
class MyV128Container
{
    public V128 v128;
    ~MyV128Container() => Console.WriteLine("Finalized MyV128Container.");
}

Span<byte> GetV128Span()
{
    var boxedV128 = new MyV128Container();
    return boxedV128.v128.AsSpan();
}

var span = GetV128Span();
GC.Collect();
GC.WaitForPendingFinalizers();
byte b1 = span[0];
```
When running on .NET 7.0 and setting a breakpoint on the `byte b1` line, the finalizer will not have run, as the `Span<byte>` tracks a field of the `MyV128Container` object, and so it is valid to access the `Span` afterwards.
However, when running it on .NET Framework, the finalizer will have run, as there the `Span<byte>` just stores a pointer value, and so accessing the `Span` outside of the `fixed` block would be invalid.

I'm not sure if there is a better solution for this (changing `V128` to a `ref struct` would cause e.g. the `V128.AllBitsSet` field to no longer work).
However, even on .NET 7.0 a similar issue can happen, when you declare a local `V128` variable (so it's on the stack), call `.AsSpan()` and return the span, as then the caller will get a span that would point to a stack area that is no longer valid.

src/Extensions.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Wasmtime
+{
+    internal static class Extensions
+    {
+        public const UnmanagedType LPUTF8Str =
+#if NETSTANDARD2_0
+            (UnmanagedType)48;
```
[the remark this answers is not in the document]
IS4Code: Yeah, I was wondering about that too. I think .NET Standard only governs the presence of a particular API, but not necessarily its full implementation, i.e. there are situations where a "supported" API actually throws `PlatformNotSupportedException`, so in my opinion it is fine to ignore this for the moment. The complete solution would probably be to copy those 6 methods that use it with a custom marshaller and use the original ones only if `(UnmanagedType)48` is actually defined at runtime. A bit complicated, but it is up to you if you feel it would be better.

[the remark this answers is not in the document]
kpreisser: Thanks! I'm OK with ignoring it for the moment.

src/V128.cs
kpreisser: > In that case, I think it would be best to just replace the usage of `GetSpan` with `fixed` where appropriate, and remove `GetSpan` from the .NET Standard 2.0 version altogether ([fcaa87f](https://github.com/bytecodealliance/wasmtime-dotnet/commit/fcaa87f7994b0234b1f6281c90fed8a4a5b345ef)).

👍 
I think it's OK to not expose a `AsSpan()` method for .NET Standard 2.0 for now. The `V128` will still be usable by using one of its constructors to set its value, and by using the `.CopyTo(Span<byte>)` method to access its value.




> The best solution is, I think, to make the reference explicit through an extension method like this one:

Good idea! With that variant, this will actually trigger the `CS8347` error in the example (I didn't know that it's possible to declare an extension method that has a `ref this` parameter, and then call it without actually specifying the `ref` keyword):

grafik

IS4Code: Yup, it was added relatively recently (only for value types), alongside `in` parameters I think. 😉

peterhuene (approved):
This looks really good. Thanks for adding this as it's definitely been a requested feature from folks.

I'll merge once @kpreisser completes their review.

kpreisser (approved):
Looks good to me too, thanks!

By the way, this actually fixes #39 which requested .NET Standard 2.0 support.

PR #259, opened 2023-06-06
Use `csharp` instead of `c#` for code blocks for the package's README.md

kpreisser:
NuGet doesn't seem to recognize `c#` as language for code blocks and doesn't apply syntax highlighting for it, whereas `csharp` seems to work - see https://www.nuget.org/packages/MySqlConnector with https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/docs/README.md for an example.

Thanks!

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -33,7 +33,7 @@ $ dotnet add package wasmtime
 
 Replace the contents of `Program.cs` with the following code:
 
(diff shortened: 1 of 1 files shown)


peterhuene (approved):
Thanks!

PR #262, opened 2023-07-03
Added `TableKind` for tables

martindevans:
Replaced `ValueKind` with `TableKind` in `Table`. This properly represents the range of type a `Table` may contain. I've minimised the amount of breakage by still accepting `ValueKind` in the constructor (marked with an `Obsolete` attribute, to steer peope towards the new constructor).

This is a small improvement (it provides a hint at valid table types) but it is still a bit fragile. It's very easy to use an incorrect `Kind` or to pass an incorrect type into `SetElement`. We may want to improve on that in the future (e.g. type safe table wrappers, like we have for globals).

### DIFF
--- a/examples/table/Program.cs
+++ b/examples/table/Program.cs
@@ -6,7 +6,7 @@ using var module = Module.FromTextFile(engine, "table.wat");
 using var linker = new Linker(engine);
 using var store = new Store(engine);
 
-var table = new Table(store, ValueKind.FuncRef, null, 4);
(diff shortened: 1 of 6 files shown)


peterhuene (commented):
Looks good!

However, there are some tests that are generating the obsolete warning by using `ValueKind` still; we should probably fix that as part of this PR.

martindevans:
Oh sorry, I must have missed those. I'll fix them now.

peterhuene (approved):

PR #280, opened 2023-10-20
Fix failing tests after the changes in bytecodealliance/wasmtime#7285

kpreisser:
Fix failing tests after the changes in bytecodealliance/wasmtime#7285.

Since that change, `threads`, `multi-memory`, and `relaxed-simd` are enabled by default. E.g. disabling `bulk-memory` also requires disabling `threads`, and disabling `simd` also requires disabling `relaxed-simd`.


peterhuene:
@kpreisser thanks for fixing this!

PR #281, opened 2023-10-24
Update to the new fuel APIs

kpreisser:
Update to the new fuel APIs introduced in bytecodealliance/wasmtime#7298.

I switched to a property so that it's possible e.g. to add or consume fuel with code like this:
```cs
store.Fuel += 100;

// Consume fuel, throwing an OverflowException if too little fuel is available.
checked {
    store.Fuel -= 150;
}
```

What do you think?
Thanks!


peterhuene (approved):
Thanks for the fix!

PR #285, opened 2023-11-03
Added support for Linker.DefineFunction with generic tuples

martindevans:
Added support for defining functions which have a tuple as a generic parameter.

Previously defining like this would work:

```csharp
Linker.DefineFunction<int, int, int>("module", "name", (int a) => (a, a));
```

But doing this would _not_ work:

```csharp
void DefineWrapper<TIn, TOut>(Func<TIn, TOut> func)
{
    Linker.DefineFunction<TIn, TOut>("module", "name", func);
}

DefineWrapper<int, (int, int)>((int a) => (a, a));
```

Even though they're ultimately doing the same thing, just with one more level of generic typing.

This PR detects that case and adds limited support for tuples (only up to 4 elements).

I have **not** added support for `ValueRaw` unboxing of tuples. As far as I can tell it's not required anywhere.

### DIFF
--- a/src/Linker.DefineFunction.cs
+++ b/src/Linker.DefineFunction.cs
@@ -8,7 +8,6 @@
 
 using System;
 using System.Buffers;
-using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Text;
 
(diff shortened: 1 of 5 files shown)


peterhuene (approved):
Thanks for the PR!

PR #289, opened 2023-12-21
`handle.IsClosed` Checks

martindevans:
Some previous relevant discussion: https://github.com/bytecodealliance/wasmtime-dotnet/pull/288#issuecomment-1861032709

The handles in wasmtime-dotnet never seem to be made invalid, so the IsInvalid checks seem to be useless. I've added `IsClosed` checks onto all the places that `IsInvalid` is used.

Previously all these paths were untested, added new tests which cover these paths.

### DIFF
--- a/src/Config.cs
+++ b/src/Config.cs
@@ -353,7 +353,7 @@ namespace Wasmtime
         {
             get
             {
-                if (handle.IsInvalid)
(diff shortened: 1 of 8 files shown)


peterhuene (approved):
Thanks for this!

PR #305, opened 2024-04-16
Fix CI.

peterhuene:
This commit:

* Updates target framework to `net8.0`.
* Fixes the `externref` API to match recent changes in upstream Wasmtime.
* Remove WASI "snapshot 0" from the tests (so old!).


peterhuene:
"The process cannot access the file 'C:\Users\runneradmin\AppData\Local\Temp\tmpbr2aoc.tmp' because it is being used by another process."

Oh, Windows, never change.

benchmarks/simple/simple.csproj
peterhuene, line 1: There's only one changed line in this file (the target framework). The rest of the file had the line endings changed from `\r\n` to `\n`.

kpreisser:
> "The process cannot access the file 'C:\Users\runneradmin\AppData\Local\Temp\tmpbr2aoc.tmp' because it is being used by another process."

FYI, on Windows it depends on which `FileShare` is specified when opening the file. When using `File.ReadAllText()`, it seems .NET specifies `FileShare.Read`, meaning opening the file fails when other code has opened the file for writing.
It should work if you manually specify `FileShare.ReadWrite`, which allows others to open the file for reading and writing, as long as they specify a `Read` file share (allowing us to open the file for reading).

guybedford (approved):

PR #308, opened 2024-06-04
Upgrade action in main.yml to no use deprecated environment

fforjan:
This should be fixing:
image


fforjan:
@peterhuene  the build are failing due to 
image

And it is independent of those changes

PR #309, opened 2024-06-04
✅enable to run unit tests in .NET 472

fforjan:
This enables to run unit test in .Net 472.

It also fixes one issue when using .Net472 - we found it when following the steps from the Readme.MD

### DIFF
--- a/src/Extensions.cs
+++ b/src/Extensions.cs
@@ -20,6 +20,7 @@ namespace Wasmtime
         {
             fixed (byte* bytesPtr = bytes)
             {
+                if(bytesPtr == null) return string.Empty;
                 return encoding.GetString(bytesPtr, bytes.Length);
             }
         }
@@ -29,6 +30,7 @@ namespace Wasmtime
         {
             fixed (byte* bytesPtr = bytes)
             {
+                if(bytesPtr == null) return string.Empty;
                 return encoding.GetString(bytesPtr, bytes.Length);
             }
         }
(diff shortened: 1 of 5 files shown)


fforjan:
Notice that build on linux only run for net80
image

While for windows, you would see:
```
Failed!  - Failed:    29, Passed:   160, Skipped:     2, Total:   191, Duration: 734 ms - Wasmtime.Tests.dll (net472)
Failed!  - Failed:    19, Passed:    80, Skipped:     0, Total:    99, Duration: 278 ms - Wasmtime.Tests.dll (net8.0)
```

jsturtevant:
@fforjan could you rebase this PR? I will take a closer look afterwards

fforjan:
@jsturtevant should be good now

fforjan:
@jsturtevant  let me know if you would like me to do anything

src/Extensions.cs
```diff
@@ -20,6 +20,7 @@ public static unsafe string GetString(this Encoding encoding, Span<byte> bytes)
         {
             fixed (byte* bytesPtr = bytes)
             {
+                if(bytesPtr == null) return string.Empty;
```
martindevans, line 23: For consistency I think this method should throw when the pointer is null. See [docs](https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=net-10.0#system-text-encoding-getstring(system-byte*-system-int32)) and [demo code](https://dotnetfiddle.net/U0TL17).

tests/WasiTests.cs
```diff
@@ -7,6 +7,54 @@
 
 namespace Wasmtime.Tests
 {
+#if NETFRAMEWORK // the following code is a polyfill as it already exists in .Net
+    using System.Runtime.CompilerServices;
+    using System.Collections;
+
+    sealed class ReferenceEqualityComparer : IEqualityComparer<object?>, IEqualityComparer
```
martindevans, line 14: Please suppress these warnings.

martindevans:
Looks like this PR needs rebasing onto master again. Thanks

Issue #315, opened 2024-06-21
CI has been failing since April


jsturtevant:
Opening this issue to start tracking down why it is failing.

The last good run was https://github.com/bytecodealliance/wasmtime-dotnet/actions/runs/8808954841 on commit  b2b0480

Started failing the next day with https://github.com/bytecodealliance/wasmtime-dotnet/actions/runs/8825026578 on commit  b2b0480

Some dependency that isn't pinned in CI might be a culprit. 

I am able to repo locally:

```
Microsoft (R) Test Execution Command Line Tool Version 17.10.0 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
The active test run was aborted. Reason: Test host process crashed : Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
Repeat 2 times:
--------------------------------
   at Wasmtime.Value+Native.wasmtime_externref_new(IntPtr, IntPtr, Finalizer)
--------------------------------
   at Wasmtime.Value.FromValueBox(Wasmtime.Store, Wasmtime.ValueBox)
   at Wasmtime.ValueBox.ToValue(Wasmtime.Store, Wasmtime.ValueKind)
   at Wasmtime.Function.Invoke(System.ReadOnlySpan`1<Wasmtime.ValueBox>)
   at Wasmtime.Function.Invoke(Wasmtime.ValueBox[])
   at Wasmtime.Tests.ExternRefTests.<ItCollectsExternRefs>g__RunTest|18_0(Int32*)
   at Wasmtime.Tests.ExternRefTests.ItCollectsExternRefs()
   at System.RuntimeMethodHandle.InvokeMethod(System.Object, Void**, System.Signature, Boolean)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(System.Object, System.Reflection.BindingFlags)
   at System.Reflection.MethodBase.Invoke(System.Object, System.Object[])
   at Xunit.Sdk.TestInvoker`1[[System.__Canon, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].CallTestMethod(System.Object)
[...]

PR #316, opened 2024-06-21
Fix Some C-API calls

martindevans:
Fixed breakage due to changes introduced in https://github.com/bytecodealliance/wasmtime/commit/e55fa3cc928547a88738597cb6236d8571c42501

Obviously I can't run all the tests, due to the other changes mentioned here https://github.com/bytecodealliance/wasmtime-dotnet/issues/315#issuecomment-2182893613. Here's what I could run:

devenv_2024-06-21_17-05-33

The WasiTests cover the changed code.

### DIFF
--- a/src/WasiConfiguration.cs
+++ b/src/WasiConfiguration.cs
@@ -454,6 +454,7 @@ namespace Wasmtime
             public static extern void wasi_config_delete(IntPtr config);
 
             [DllImport(Engine.LibraryName)]
+            [return: MarshalAs(UnmanagedType.I1)]
             public unsafe static extern bool wasi_config_set_argv(Handle config, nuint argc, byte** argv);
 
             [DllImport(Engine.LibraryName)]
--- a/src/WasiConfiguration.cs
+++ b/src/WasiConfiguration.cs
@@ -299,7 +299,9 @@ namespace Wasmtime
             {
                 fixed (byte** arrayOfStringsPtrNamedArgs = args)
                 {
-                    Native.wasi_config_set_argv(config, _args.Count, arrayOfStringsPtrNamedArgs);
+                    // This shouldn't ever fail - it would only return false if `ToUTF8PtrArray` creates invalid UTF8 bytes!
+                    if (!Native.wasi_config_set_argv(config, (nuint)_args.Count, arrayOfStringsPtrNamedArgs))
+                        throw new WasmtimeException("Failed to encode string to UTF8");
                 }
             }
             finally
@@ -329,7 +331,9 @@ namespace Wasmtime
 
             try
             {
-                Native.wasi_config_set_env(config, _vars.Count, names, values);
+                // This shouldn't ever fail - it would only return false if `ToUTF8PtrArray` creates invalid UTF8 bytes!
+                if (!Native.wasi_config_set_env(config, (nuint)_vars.Count, names, values))
+                    throw new WasmtimeException("Failed to encode string to UTF8");
             }
             finally
             {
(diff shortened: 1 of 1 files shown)


jsturtevant:
Thanks!  I've opened #317  which should help some.  Maybe we need both?

jsturtevant:
Looking alittle closer this, seem to only be a change in the wasmtime 0.22.0 versions. We will also need https://github.com/bytecodealliance/wasmtime/commit/77405cc8fdfb12e5a6022a338ad371d7cf357687. 

I am guessing we are missing other API changes between 0.19 and 0.22 which is why when pulling this into #317  didn't work

martindevans:
Yeah this is only fixing the first of the two things you linked. We'll need another PR to fix https://github.com/bytecodealliance/wasmtime/commit/77405cc8fdfb12e5a6022a338ad371d7cf357687.

At the end I guess we'll need to merge all 3 PRs together to test it it finally fixes all of the tests.

jsturtevant:
> At the end I guess we'll need to merge all 3 PRs together to test it it finally fixes all of the tests.

Yes, I would like to group the changes to get CI passing in order to merge.  Without that I am not confident enough in the changes.  I won't be able to get to https://github.com/bytecodealliance/wasmtime/commit/77405cc8fdfb12e5a6022a338ad371d7cf357687 today if you are interested in picking that up too.  Thanks for the help!

martindevans:
Unfortunately I don't think I'll have time today or tomorrow. It looks like a pretty complex set of changes!

src/WasiConfiguration.cs
```diff
@@ -450,12 +454,12 @@ private static class Native
             public static extern void wasi_config_delete(IntPtr config);
 
             [DllImport(Engine.LibraryName)]
-            public unsafe static extern void wasi_config_set_argv(Handle config, int argc, byte** argv);
+            public unsafe static extern bool wasi_config_set_argv(Handle config, nuint argc, byte** argv);
```
kpreisser, line 457: I think this is missing the `[return: MarshalAs(UnmanagedType.I1)]` attribute (see #170).
```suggestion
            [return: MarshalAs(UnmanagedType.I1)]
            public unsafe static extern bool wasi_config_set_argv(Handle config, nuint argc, byte** argv);
```

src/WasiConfiguration.cs
```diff
@@ -450,12 +454,12 @@ private static class Native
...
 
             [DllImport(Engine.LibraryName)]
-            public static extern unsafe void wasi_config_set_env(
+            public static extern unsafe bool wasi_config_set_env(
```
kpreisser, line 460: Same here:
```suggestion
            [return: MarshalAs(UnmanagedType.I1)]
            public static extern unsafe bool wasi_config_set_env(
```

kpreisser (commented):
Thanks for updating this!

It looks like the `MarshalAs(UnmanagedType.I1)` attribute is missing from the function delcaration, as without this, .NET would marshal the `bool` value as 4-byte `BOOL`, rather than as 1 byte value.

Otherwise, this looks good to me.

kpreisser (approved):

jsturtevant:
@martindevans could you rebase? then we can get this in for 1.22 release

martindevans:
@jsturtevant I'm on holiday at the moment. I'll be able to look at it on Sunday at the soonest. 

If you want to take these changes and incorporate them yourself before then go ahead :)

martindevans:
I just discovered GitHub itself has a button to automatically do the rebase, so I don't need access to my laptop after all!

jsturtevant (approved):

jsturtevant:
> @jsturtevant I'm on holiday at the moment. I'll be able to look at it on Sunday at the soonest.

no worries. Thanks and I hope you I didn't interrupt your personal time, that always takes precedent

PR #336, opened 2025-07-11
Replace Marshal with Unsafe

andreakarasho:


andreakarasho:
Closing this because seems to be necessary anymore.

---- 
Side note:
> I noticed the [warning](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafe.sizeof?view=net-9.0#remarks) in the documentation. It seems to me that we actually want the size of the _unmanaged_ view of the type? In that case we would need to use [Marshal.SizeOf](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.sizeof?view=net-9.0), as it is now.

Seems like they updated the documentation and removed that warning (?). MS now suggets to use `sizeof` (`Unsafe.SizeOf` is just a wrapper around `sizeof`) for blittable structs (which this was the case).
https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices#structs

PR #339, opened 2025-07-14
Added `Wasi2Configuration`

martindevans:
Added `Wasi2Configuration` which mirrors `WasiConfiguration` for WASI preview2.

### DIFF
--- a/tests/Wasi2Tests.cs
+++ b/tests/Wasi2Tests.cs
@@ -1,0 +1,18 @@
+﻿using Xunit;
+
+namespace Wasmtime.Tests;
+
+public class Wasi2Tests
+{
+    [Fact]
+    public void ItSetsWasi2Config()
+    {
+        using var engine = new Engine();
+        using var store  = new Store(engine);
+
+        var config = new Wasi2Configuration();
+        config.WithInheritedStandardInput();
+
+        store.SetWasiConfiguration(config);
+    }
+}
--- a/tests/Wasi2Tests.cs
+++ b/tests/Wasi2Tests.cs
@@ -1,4 +1,5 @@
-﻿using Xunit;
+﻿using System;
+using Xunit;
 
 namespace Wasmtime.Tests;
 
@@ -12,6 +13,8 @@ public class Wasi2Tests
 
         var config = new Wasi2Configuration();
         config.WithInheritedStandardInput();
+        config.WithInheritedStandardOutput();
+        config.WithInheritedStandardError();
 
         store.SetWasiConfiguration(config);
     }
--- a/src/Store.cs
+++ b/src/Store.cs
@@ -72,6 +72,13 @@ namespace Wasmtime
             }
         }
 
+        internal void SetWasiConfiguration(Wasi2Configuration config)
+        {
+            var wasi = config.Build();
+            Native.wasmtime_context_set_wasip2(handle, wasi.DangerousGetHandle());
+            wasi.SetHandleAsInvalid();
+        }
+
         /// <summary>
         /// Configures the relative deadline at which point WebAssembly code will trap.
         /// </summary>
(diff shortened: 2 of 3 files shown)


tests/Wasi2Tests.cs
```diff
@@ -0,0 +1,18 @@
...
+public class Wasi2Tests
+{
+    [Fact]
+    public void ItSetsWasi2Config()
+    {
+        using var engine = new Engine();
+        using var store  = new Store(engine);
+
+        var config = new Wasi2Configuration();
+        config.WithInheritedStandardInput();
```
jsturtevant, line 15: Any reason to not test the other configs?

jsturtevant (commented):
LGTM

@kpreisser @BasGeertsema want to give a review?

tests/Wasi2Tests.cs
martindevans: I'll have a look at adding some more tests later today

martindevans: I have added some more coverage (just calling other `WithInheritedX` methods).

I'm not really happy about these tests though - they don't actually assert anything!

Ideally I guess we'd need a `WASI2.wat` test file that reads some input, logs some out and logs some error output. Then we could assert that all of that is written to the correct outputs.

BasGeertsema: I looked around for an easy way to test this but could not find it and I wonder what the right approach is to test this.

The [WASI preview 2](https://github.com/WebAssembly/WASI/blob/main/wasip2/README.md) is quite an API surface. In the [stdio test in wasi-testsuite](https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/stdio.rs) a lot more functionality is tested than just stdin/stdout/stderr. On the surface, it seems difficult to test _just_ the stdin/stdout in isolation with a simple .wat file. For example, the [WASI 0.2 WIT for the CLI](https://github.com/WebAssembly/wasi-cli/blob/main/wit/stdio.wit)  contains references to the `output-stream` and `input-stream` types and other WASI specs. The WASI 0.2 tests in the test-suite are compiled to `.wasm` via build scripts and I have not yet seen a simple stable `.wat` file that can only test stdin/stdout.

This however got me thinking what the proposition is of `wasmtime-dotnet`. The `wasmtime` project links to this project as a way to embed `wasmtime` as a library. I guess historically the use case was primarily around embedding an engine capable of _executing_ WASM modules. But with the advent of WASI, `wasmtime` offers a runtime with rich functionality that can interact with its environment through these system interfaces. So, whole applications (CLI applications and web applications for example) can be compiled to wasm and executed directly by invoking `wasmtime` with the wasm file as argument. Kind of like an application server.

In contrast, `wasmtime-dotnet` is a library (NuGet package) and does _not_ have a standalone executable. So it is not possible to directly execute a WASM application, rather you´d have to write the dotnet program yourself that uses `wasmtime-dotnet`.

If `wasmtime-dotnet` _would_ offer such an executable, we could utilize the [wasi-testsuite](https://github.com/WebAssembly/wasi-testsuite) to test the whole WASI surface. Which would cover the stdin/stdout functionality of this pull request, and a whole lot more. But at the same time it would be ´just' a thin wrapper around `wasmtime` where all of the WASI interfaces are implemented by the `wasmtime` runtime. Which is not a very good use case, because why not directly use `wasmtime` in that case?

Imho the value proposition of `wasmtime-dotnet` is that .NET applications can enhance their applications with WASM modules/components (possibly user-supplied) in a safe way that feels native to .NET (i.e. ergonomic to use). But that also implies more control over the system interfaces. For example, if a WASM component is loaded that uses [wasi-cli](https://github.com/WebAssembly/wasi-cli) or [wasi-sockets](https://github.com/WebAssembly/wasi-sockets) do we want to use .NET streams and sockets or it is offloaded completely to the wasmtime (Rust) runtime, over which the .NET application has very little control? Integration with .NET streams and sockets would be ideal imo, but is also a _huge_ endeavour.

I'm relatively new to the WASM ecosystem so I might be very confused here but imo there should be a clear proposition for wasmtime-dotnet that can guide the development of wasmtime-dotnet, which would also direct questions such as how to test WASI functionality.

PR #346, opened 2025-08-17
Component Support WIP

martindevans:
This is a rough first pass looking at what's required for wasm component support. I've pulled in everything from the `component` folder of the C-API (see [here](https://docs.wasmtime.dev/c-api/dir_788242682a7aad0d494995df051a8421.html)) and have wrapped some of the easiest bits in the normal boilerplate (creating handles etc).

I'm absolutely **not** intending to complete this work solo!

We can use this branch to gather up any component related bits until we have something that works and it worth merging into `main`.

### DIFF
--- a/src/Components/ComponentValue.cs
+++ b/src/Components/ComponentValue.cs
@@ -1,0 +1,130 @@
+﻿using System.Drawing;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using static System.Runtime.InteropServices.JavaScript.JSType;
+
+namespace Wasmtime.Components;
+
+// todo: everything here: https://docs.wasmtime.dev/c-api/component_2val_8h.html
+/*
+ wasmtime_component_vallist
+ A vec of a struct wasmtime_component_val
+
+ wasmtime_component_valrecord
+ A vec of a struct wasmtime_component_valrecord_entry
+
+ wasmtime_component_valtuple
+ A vec of a struct wasmtime_component_val
+
+ wasmtime_component_valflags
+ A vec of a wasm_name_t
+
+ wasmtime_component_valvariant_t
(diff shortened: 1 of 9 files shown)


src/Components/ComponentValue.cs
```diff
@@ -0,0 +1,130 @@
+﻿using System.Drawing;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using static System.Runtime.InteropServices.JavaScript.JSType;
```
jsturtevant: What's this used for?  Seems like its a browswer only API https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.javascript.jstype?view=net-9.0

martindevans: I don't think that is actually used, not sure how/why it got auto added to the `using`s list!

martindevans: Removed.

cyraid (commented):
Is there a reason for making these on to a single line vs multi-line? The multi-line is a bit more readable. :)

UserMist:
I am currently making wasmtime bindings for Beef, language which is basically lower level C#, and I expect to mostly finish them in the next week or two. Once that's done, you can use my work as reference.

phillipcaofph:
Hi, just wanting to know if there is going to be any progress with this PR? I'm keen to have component support via this package if possible

martindevans:
I'm not personally planning to work on it anytime soon, but I will be very happy to merge PRs towards component support!

phillipcaofph:
@martindevans Shall I target PRs into your fork? Given there is progress there already

martindevans:
Yes please. Let's use this as our branch to develop component features, until we have something worth merging into master.

PR #355, opened 2026-02-01
Fix Windows interop and toggle dev ABI with DevBuild

tomohisa:
## Summary
  - Use cdecl for wasmtime P/Invoke and native callbacks.
  - Toggle dev/stable ABI layouts and TrapCode by DevBuild.
  - Document DevBuild switch in README.

## Why
CI runs with DevBuild=true (wasmtime-dev), which requires dev ABI layouts.
This fixes test-host crashes such as `unknown wasmtime_valkind_t`.

  ## Tests
  - (CI) `dotnet test Wasmtime.sln`

### DIFF
--- a/src/TrapException.cs
+++ b/src/TrapException.cs
@@ -37,8 +37,112 @@ namespace Wasmtime
         Unreachable = 9,
         /// <summary>The trap was the result of interrupting execution.</summary>
         Interrupt = 10,
+#if WASMTIME_DEV
+        /// <summary>The trap was the result of running out of the configured fuel amount.</summary>
+        OutOfFuel = 11,
         /// <summary>
-        /// The trap was the result of executing a function that was `canon lift`'d, then `canonlower`'d, then called. 
+        /// The trap was the result of atomic wait operations on non-shared memory.
+        /// </summary>
+        AtomicWaitNonSharedMemory = 12,
+        /// <summary>
+        /// The trap was the result of a call to a null reference.
+        /// </summary>
+        NullReference = 13,
+        /// <summary>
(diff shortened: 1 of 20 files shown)


src/TrapException.cs
```diff
@@ -37,8 +37,112 @@ public enum TrapCode
         Unreachable = 9,
         /// <summary>The trap was the result of interrupting execution.</summary>
         Interrupt = 10,
+#if WASMTIME_DEV
```
martindevans, line 40: I can't see any mention in the wasmtime docs that these traps only exist in dev builds. Are you certain this is correct?

tomohisa: Thanks for checking. I agree this isn’t about ‘dev‑only’ traps. The issue is ABI layout: the C API enum in trap.h lists entries without explicit numeric assignments, so the ordinal values are defined by order and can shift when new variants are inserted. [docs.wasmtime.dev](https://docs.wasmtime.dev/c-api/trap_8h_source.html)

The Rust docs also note that C‑API values and Rust trap codes must be kept in sync and that the enum is non‑exhaustive. [rustdocs.bsx.fi](https://rustdocs.bsx.fi/wasmtime/enum.TrapCode.html) That’s why #if WASMTIME_DEV currently selects the ABI layout that matches the Wasmtime build being linked. I’m open to making this more explicit (e.g., WASMTIME_TARGET_VERSION=35 vs =dev), but the immediate goal is to avoid ABI mismatches so latest builds don’t fail by default.

jsturtevant:
I don't think this is needed after https://github.com/bytecodealliance/wasmtime-dotnet/pull/361. Feel free to send a sperate PR if you find something is still needed.

PR #359, opened 2026-04-11
Update to Wasmtime v44.0.0

renshyle:


martindevans (approved):

jsturtevant (approved):

src/Value.cs
```diff
@@ -531,6 +523,8 @@ internal struct AnyRef
         private uint __private1;
 
         private uint __private2;
+
+        private uint __private3;
```
kpreisser, line 527: Note that `__private3` in `wasmtime_anyref` is declared in `val.h` as follows:
```c
  /// Internal to Wasmtime.
  void *__private3;
```

Therefore, this should actually be `IntPtr` (pointer-sized) instead of `uint` (32-bit).

```suggestion
        private IntPtr __private3;
```

src/Value.cs
```diff
@@ -541,5 +535,7 @@ internal struct ExternRef
...
```
kpreisser, line 539: Same here: `__private3` in `wasmtime_externref` is defined as:
```c
  /// Internal to Wasmtime.
  void *__private3;
```


```suggestion
        private IntPtr __private3;
```

kpreisser (commented):
Hi, sorry for being late. The changes look good, except that the data type for the new `__private3` fields seems to be the wrong one.

src/Value.cs
martindevans: Well spotted. I've open a PR here https://github.com/bytecodealliance/wasmtime-dotnet/pull/359 to fix it.

src/Value.cs
martindevans: Well spotted. I've open a PR here https://github.com/bytecodealliance/wasmtime-dotnet/pull/359 to fix it.

martindevans:
@renshyle Did you look at v45 while working on this PR? I've been investigating the v45 update and have encountered a lot of issues.

Issue #324, opened 2024-07-09
Component model support


JanProvaznik:
### Context 
[https://github.com/bytecodealliance/wasmtime/tree/main/examples/component](https://github.com/bytecodealliance/wasmtime/tree/main/examples/component)
Add support for making the same example from C#.

[https://github.com/WebAssembly/component-model/tree/main](https://github.com/WebAssembly/component-model/tree/main)
[https://github.com/bytecodealliance/componentize-dotnet](https://github.com/bytecodealliance/componentize-dotnet)

### Current state:
Only Wasm modules can be instantiated and run functions from them. Loading .wasm file which is compiled as a component into a module fails.

### Steps
[ ] running components
[ ] imports and exports (with WIT data types)
[ ] converting modules to components
[ ] multiple components interacting

### Remarks
Unsure how complex implementing this is, likely nontrivial since components should support WIT.

jsturtevant:
Now that we released wasmtime v22 support, we should have the right version but may need to expose some new apis

jsturtevant:
Since we use the c-api we need https://github.com/bytecodealliance/wasmtime/issues/8036.

BasGeertsema:
I'm also very much interested in supporting Wasm Component.

The [33 release](https://github.com/bytecodealliance/wasmtime/releases/tag/v33.0.0) and [34 release](https://github.com/bytecodealliance/wasmtime/releases/tag/v34.0.0) expanded the C api to support (some aspects of) Wasm Components.

This is encouraging and it seems we can make some progress here as well?

jsturtevant:
Yes, the project could begin to consume those.  Its does need a refresh to pick up the latest wasmtime versions

BasGeertsema:
> Yes, the project could begin to consume those. Its does need a refresh to pick up the latest wasmtime versions

I have created a PR for the refresh: #332

cyraid:
Good job on the refresh btw!

Thinking about the component model, how would it work? Could we dress up some classes with attributes like Microsoft's clearscript?

BasGeertsema:
> Good job on the refresh btw!
> 
> Thinking about the component model, how would it work? Could we dress up some classes with attributes like Microsoft's clearscript?

[...]

PR #360, opened 2026-05-03
feat(components): WASI 0.2 host bindings + Roslyn source generator

kitsunoff:
## Summary

Implements WASI 0.2 Component Model host-side support for wasmtime-dotnet, addressing #324.

Two layers:

1. **Runtime API** (`Wasmtime.Components`) — `SafeHandle`-backed wrappers around `wasmtime_component_*` C API: `Component`, `ComponentLinker`, `ComponentLinkerInstance`, `ComponentInstance`, `ComponentFunction`, `ComponentValue` (discriminated union for every WASI 0.2 value kind: bool / integers / floats / char / string / list / record / tuple / variant / enum / option / result / flags). Host-defined imports via `ComponentLinkerInstance.DefineFunc` with idiomatic `(ReadOnlySpan<ComponentValue>, Span<ComponentValue>)` callback.

2. **Roslyn source generator** (`Wasmtime.Component.SourceGenerators`) — turns WIT files into idiomatic C# bindings. Consumes JSON IR from `wasm-tools component wit X.wit --json` instead of a hand-rolled WIT parser. Emits strongly-typed records / enums / flags / variants nested under the user's `[ComponentBindings]` partial class, per-export call wrappers, and an `IImports` interface plus static `RegisterImports(linker, impl)` for world imports.

End-to-end fixture (`tests/Components/fixtures-src/`) is a .NET component built by \`componentize-dotnet\` (NativeAOT-LLVM) — proves bidirectional .NET host ↔ .NET component for every composite kind, including a host-defined \`host-double\` import the component calls back into.

Documentation in \`docs/component-model.md\` plus a new README section. Follow-up items tracked in \`docs/component-model-followups.md\`.

## Test plan

- [x] Existing core wasm tests still pass on netstandard2.0 / netstandard2.1 / net8.0 / net9.0
- [x] \`ComponentValue\` layout verified at runtime (\`Marshal.SizeOf == 32\`); same for \`wasmtime_component_func_t\` (24 bytes — Rust uses an anonymous nested struct with trailing padding, mismatch caused entire-class of wrong-function-returned bugs during development)
- [x] End-to-end tests load and call cargo-component / componentize-dotnet built \`.wasm\` components covering every composite kind
- [x] Host-defined imports invoked from inside the component, plus host-throwing-exception → wasmtime trap propagation
- [x] Generator unit tests cover JSON IR parsing, type emission for record/enum/flags/variant, primitive + composite call wrappers, import registration

## Known limitations (tracked in \`docs/component-model-followups.md\`)

- **WIT \`resource\` types not supported.** wasmtime C API gained the resource surface only in v42; this branch keeps v35. Standard WASI 0.2 interfaces that use resources internally still work because wasmtime's native impl handles those tables — limitation is custom WIT resources that cross the managed boundary
- **Async types** (\`stream\`, \`future\`, \`error-context\`) — WASI 0.3, out of scope
- **Memory ownership** — \`ComponentValue.ownsHeap\` byte squats Rust enum padding; safe in practice today but technically UB. Refactor to explicit per-call ownership tracking is in follow-ups
- **\`post_return\` ordering** — currently invoked inside \`Call\` before the caller reads results; relies on wasmtime's eager-clone implementation detail
- **Generator edge cases** — \`option<tuple<...>>\` and named type aliases (\`type x = list<u32>\`) need additional emit logic

## Building the wasm fixture

NativeAOT-LLVM has no macOS prebuilts, so the fixture builds inside an arm64 Linux container:

\`\`\`bash
./tests/Components/regenerate.sh
\`\`\`

Pre-built \`.wasm\` and \`.wit.json\` artifacts are committed so consumers don't need the Linux toolchain.

## Credits

Builds on the sketch in #346 by @martindevans — same \`Wasmtime.Components\` namespace and SafeHandle pattern, with the rest of the C API surface, marshalling, source generator, and end-to-end tests added.


kitsunoff:
CI panic (\`unknown wasmtime_valkind_t: 226\` at \`c-api/src/val.rs:392\`) is the predicted manifestation of follow-up #1/#2 documented in \`docs/component-model-followups.md\`: \`ComponentValue.ownsHeap\` squats on Rust's \`#[repr(C, u8)]\` enum padding, which is \`MaybeUninit\` per the Rust reference. In Release builds the padding bytes carry non-zero garbage often enough that \`Free()\` switches into composite cleanup paths and runs \`Marshal.FreeHGlobal\` on Rust-allocated pointers — heap corruption then surfaces in adjacent core wasm tests as a bogus valkind read.

The fix shape (split \`Free()\` into \`FreeManaged()\` for managed-allocated values and \`ReleaseRustOwned\` calling \`wasmtime_component_val_delete\` for wasmtime-filled slots, with ownership tracked outside the C ABI footprint) is already in the follow-ups doc. I started this refactor on this branch but the combined diff with the \`post_return\`-after-lift change (#3) tripped a separate \`panic!("None")\` in \`c-api/src/store.rs:116:30\` that needs a smaller upstream repro. Rolled back to keep the rest of the work mergeable, will land it in a follow-up PR before marking this ready for review.

Other follow-ups (#4 \`option<tuple<...>>\`, #5 type aliases, #6 csproj duplicate) are simpler and can ride the same follow-up.

jsturtevant:
Thanks for you interest in contributing to the project.  This looks to be mostly AI generated and is a lot of code.  There was also some progress made towards this in https://github.com/bytecodealliance/wasmtime-dotnet/pull/346

If you would like to go forward with this maybe the next steps are to break this down into small chunks and build off and collaborate with the discussion in https://github.com/bytecodealliance/wasmtime-dotnet/issues/324?

kitsunoff:
> Thanks for you interest in contributing to the project. This looks to be mostly AI generated and is a lot of code. There was also some progress made towards this in #346
> 
> If you would like to go forward with this maybe the next steps are to break this down into small chunks and build off and collaborate with the discussion in #324?

Yes, of course!

I simply implemented everything using the AI ​​agent as a large pull request because I was doing it literally as a quick proof of concept.
And it definitely needs polishing and a lot of testing.
Partially it's works, but not all.
That's why there's a lot of code, and some of it isn't very good.

jsturtevant:
Great, Thanks!  It can be very useful to do that type of exploration.  To get it merged we will need to break it down into manageable, reviewavle chunks.  There has been a lot of interest in components so I know there will be excitement if we can do that.

This is useful as a reference as well https://github.com/bytecodealliance/governance/blob/main/AI_TOOL_POLICY.md

jsturtevant:
I am going to close this but feel free to open up smaller targeted prs

PR #362, opened 2026-05-28
WASM Exceptions And Tag Import/Export

martindevans:
- Added support for WASM tag types (import/export)
 - Added support for enabling WASM exceptions

This adds the ability to load modules which depend upon WASM exceptions. It does *not* add any support for catching, throwing or interacting with exceptions in any way (still working on that for a follow-up PR).

### DIFF
--- a/src/Config.cs
+++ b/src/Config.cs
@@ -258,13 +258,24 @@ namespace Wasmtime
         }
 
         /// <summary>
+        /// Configures whether the WebAssembly exceptions proposal is enabled.
+        /// </summary>
+        /// <param name="enable">True to enable exceptions or false to disable.</param>
+        /// <returns>Returns the current config.</returns>
+        public Config WithExceptions(bool enable)
+        {
+            Native.wasmtime_config_wasm_exceptions_set(handle, enable);
+            return this;
+        }
+
+        /// <summary>
         /// Sets whether the WebAssembly stack switching proposal is enabled. 
         /// </summary>
         /// <param name="enable">True to enable WebAssembly stack switching support or false to disable.</param>
         /// <returns>Returns the current config.</returns>
         private Config WithStackSwitching(bool enable)
         {
-            // todo: unlikely to be compatible with wasmtime-dotnet on Windows due to threads
+            // warning: **unlikely to be ever compatible with wasmtime-dotnet on Windows due to fibers**
 
             Native.wasmtime_config_wasm_stack_switching_set(handle, enable);
             return this;
@@ -631,6 +642,9 @@ namespace Wasmtime
 
             [DllImport(Engine.LibraryName)]
             public static extern void wasmtime_config_wasm_component_model_set(Handle config, [MarshalAs(UnmanagedType.I1)] bool value);
+
+            [DllImport(Engine.LibraryName)]
+            public static extern void wasmtime_config_wasm_exceptions_set(Handle config, [MarshalAs(UnmanagedType.I1)] bool value);
 
             // todo: void wasmtime_config_host_memory_creator_set(wasm_config_t *, wasmtime_memory_creator_t *)
         }
--- a/src/Export.cs
+++ b/src/Export.cs
@@ -23,19 +23,20 @@ namespace Wasmtime
             }
 
             var exports = new Export[(int)this.size];
-            for (int i = 0; i < (int)this.size; ++i)
+            for (var i = 0; i < (int)this.size; ++i)
             {
                 var exportType = this.data[i];
                 var externType = Native.wasm_exporttype_type(exportType);
 
-                exports[i] = (WasmExternKind)Native.wasm_externtype_kind(externType) switch
+                var kind = (WasmExternKind)Native.wasm_externtype_kind(externType);
+                exports[i] = kind switch
                 {
                     WasmExternKind.Func   => new FunctionExport(exportType, externType),
                     WasmExternKind.Global => new GlobalExport(exportType, externType),
                     WasmExternKind.Table  => new TableExport(exportType, externType),
                     WasmExternKind.Memory => new MemoryExport(exportType, externType),
-
-                    _ => throw new NotSupportedException("Unsupported export extern type.")
+                    WasmExternKind.Tag    => new TagExport(exportType, externType),
+                    _ => throw new NotSupportedException($"Unsupported export extern type: {kind}.")
                 };
             }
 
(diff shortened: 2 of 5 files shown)


Avril112113 (approved):

Avril112113:
I've been working on adding proper exceptions handling along with @martindevans (for a separate PR)
I've tested the tag import and exports in a project of mine, besides the tag `Parameters`, but the tests cover those anyway.

From prior testing, enabling exceptions in the config is no issue if there is no C# call boundary.
However, if there is a C# call boundary, a trap is triggered which boils all the way down in C# only, skipping WASM entirely.
So until the next PR, it's worth noting as it could be odd behaviour for WASM expecting to catch (for example, with setjmp/longjmp).

jsturtevant:
Is the next pr in flight?  Should we wait on this or is it ready to go?

kpreisser:
Hi!

> However, if there is a C# call boundary, a trap is triggered which boils all the way down in C# only, skipping WASM entirely.

I haven't look into how WASM exceptions work exactly, but I assume the issue here is that WASM exceptions will be wrapped in a `wasm_trap_t*` when a WASM function call returns, and for each host-to-wasm transition in `wasmtime-dotnet` (WASM function calls), whenever a trap occurs in WASM, it is converted into a `TrapException` by [`TrapException.FromOwnedTrap`](https://github.com/bytecodealliance/wasmtime-dotnet/blob/43538a8e4f395b5fd2e8e27361c1e0e1d8b10aa6/src/TrapException.cs#L407-L419), which is then thrown and the `wasm_trap_t*` is deleted immediately.
Howver, if there is a wasm-to-host transition (**callback**) on the stack, it will catch exceptions in [`Function.HandleCallbackException()`](https://github.com/bytecodealliance/wasmtime-dotnet/blob/43538a8e4f395b5fd2e8e27361c1e0e1d8b10aa6/src/Function.cs#L696) and create a **new `wasm_trap_t*`** for it, so the original trap is not preserved, even if the user stack frames (.NET methods) within these transitions don't actually catch the exception (so the WASM exception wrapped in a trap would be lost).

Do you know if this the case, that the `wasm_trap_t*` would need to preserved between function calls and callbacks, to retain the WASM exception if the .NET code in between doesn't want to catch it?

If yes, then in order to preserve the trap between host-to-wasm and wasm-to-host transitions, a possibility I could imagine is to use a similar mechanism that we already use for preserving a .NET exception between a wasm-to-host and host-to-wasm transition, which is done by storing the .NET exception in a **static thread-local** [`CallbackErrorCause` field](https://github.com/bytecodealliance/wasmtime-dotnet/blob/43538a8e4f395b5fd2e8e27361c1e0e1d8b10aa6/src/Function.cs#L791-L792) when it's caught at the wasm-to-host transition, and then using it as `InnerException` on the next host-to-wasm transition when throwing a new `WasmtimeException` due to an `wasmtime_error_t*`.

For example, we could add a static thread-local counter field that counts the number of wasm-to-host transitions (callbacks) currently on the stack, as well as a static thread-local field that can store a `TrapException` and a `wasm_trap_t*`.

Now, when a WASM function returns a trap and we call `TrapException.FromOwnedTrap`, we could check if the counter is > 0, in which case we throw the exception but specify `delete: false` to not delete the trap pointer; instead, we store the thrown `TrapException` together with the trap pointer in the static thread-local field.
Then, when an exception is caught at the next callback (wasm-to-host transition), it can check whether the caught exception is the same instance as stored in the thread-local static field, and in that case, just return the stored trap pointer; otherwise, when it's a different exception instance, delete the stored trap pointer and then create a new trap for the exception and return that instead. (Or, if no exception is caught at the callback, it would still need to delete the trap pointer if present.)

What do you think?

---

By the way, I remember that some time ago, the trap handling in upstream Wasmtime was changed, so that **host-defined traps** (with `wasmtime_trap_new` and returned as `wasm_trap_t*` from a callback) are then returned as `wasmtime_error_t*` from the function call (see #175), do you know whether this is still the case?
(Because that would seem like a discrepancy between simple host-defined `wasm_trap_t*` that are then returned as `wasmtime_error_t*` from the function call, and exception-wrapping traps that are returned as `wasm_trap_t*`.)

If not, the logic in `TrapException.FromOwnedTrap()` would probably need to be adjusted to contain the same code as in [`WasmtimeException.FromOwnedError`](https://github.com/bytecodealliance/wasmtime-dotnet/blob/43538a8e4f395b5fd2e8e27361c1e0e1d8b10aa6/src/WasmtimeException.cs#L40-L48), so that it checks and clears the `CallbackErrorClause` and uses it as `InnerException` for the `TrapException`.

Thanks!

martindevans:
> Is the next pr in flight? Should we wait on this or is it ready to go?

I'd like to get this one merged first. Full support for exceptions is _probably_ blocked on v45, so I'm going to try and pull out self contained parts into separate PRs before then.

In it's current state this is more useful than it sounds. This is sufficient for any module that uses exceptions without letting them cross the C# call boundary, I would guess that's probably the majority of them.

> I haven't look into how WASM exceptions work exactly, but I assume the issue here is that WASM exceptions will be wrapped in a wasm_trap_t*

It does return a trap, but the trap is kind of a dummy trap object that indicates an exception happened, you can call `wasmtime_context_has_exception`/`wasmtime_context_take_exception` ([docs](https://docs.wasmtime.dev/c-api/exnref_8h.html#adda9ab17968068f565e2454db79b5ecc)) to get the actual exception info. My plan is to add another exception type (`WasmException`) which holds the exception data.

> Do you know if this the case, that the wasm_trap_t* would need to preserved between function calls and callbacks, to retain the WASM exception if the .NET code in between doesn't want to catch it?

I haven't looked at it in detail yet (@Avril112113 might know more) but yes I think we'll have take ownership of the exception when entering C# (`wasmtime_context_take_exception`), then on transitioning back to wasm pass the exception back (`wasmtime_context_set_exception`).

> Call depth counter

That's pretty much what I came up with too 👍

Avril112113:
@kpreisser Regarding your first section, I haven't looked into exactly why the exception isn't being passed through the C# boundary, but if the original trap isn't being rethrown, but instead a brand new one, that'd be a healthy guess.
This gist is an example of how they work right now (and with this PR); https://gist.github.com/Avril112113/f62593075509b56a8e5940a84c72696c

> Do you know if this the case, that the `wasm_trap_t*` would need to preserved between function calls and callbacks, to retain the WASM exception if the .NET code in between doesn't want to catch it?

I think the trap can be dropped if we're handling it as an exception instead of a normal trap;
We take (including ownership of) the exception, if it's within context during a trap via `wasmtime_context_take_exception`.
If we are re-throwing it, then `wasmtime_context_set_exception` takes (including ownership of) the exception and returns a new trap.
If we don't intend to pass it back (ie, user handles it), we will need to clean-up the exception object.
*In current testing, we are not yet handling clean-up of the exception or trap.*

> Is the next pr in flight? Should we wait on this or is it ready to go?

Handling of the exception object seems to have changed between v44 and v45, as they changed from `exn` type to `exnref` in v45.
This PR at least allows WASM to use exceptions without much concern about v45 changes.
The only notable caveat, is WASM can't use them across the host call boundary (see gist above for example).

jsturtevant (approved):

martindevans:
Merge conflicts have been resolved.

jsturtevant (approved):

PR #369, opened 2026-06-12
Use trusted publishing for NuGet releases

jsturtevant:
NuGet releases can avoid relying on a long-lived API key secret by using GitHub OIDC to request a short-lived NuGet API key at publish time.

This updates the publish workflow to grant `id-token: write`, authenticate with `NuGet/login` pinned to the full commit hash, and use the action output when pushing the generated package.

https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing

note: unfortunately you have to use your own username but the trusted publishing registration owner is bytecodealliance

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -8,6 +8,10 @@ on:
     tags:
     - 'v*'
 
+permissions:
+  contents: read
+  id-token: write
+
 jobs:
   publish:
     name: "Publish NuGet Package"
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -24,7 +28,12 @@ jobs:
       run: |
         cd src
         dotnet pack -c Release /p:Packing=true
+    - name: NuGet login
+      uses: NuGet/login@8d196754b4036150537f80ac539e15c2f1028841
```
martindevans, line 32: Is there a reason to use this exact version instead of `v1.2.0` (latest)?

martindevans: Oh I see, this is that version but pinned to the ID for safety.

martindevans (approved):

.github/workflows/publish.yml
jsturtevant: yea, its generally the recommended thing these days https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions even if its not third party

martindevans: I think this PR and #370 have given me a **lot** of work to do on 50 other repos of mine over the weekend! 😅

jsturtevant: the nice thing is that dependabot works with sha's once you've moved over!

PR #373, opened 2026-08-16
Add epoch deadline callback binding to Store

rlauuzo:
Wasmtime's `wasmtime_store_epoch_deadline_callback` was unbound, so the only way to bound wasm execution was a single `SetEpochDeadline` fixed at call start. Stopping a running guest early meant incrementing the engine-wide epoch, which traps every other in-flight store sharing the engine.

`Store.SetEpochDeadlineCallback` binds it: the callback runs on the executing thread when the deadline is reached, and either returns a new relative deadline to resume with, or throws to terminate the execution. A throwing callback surfaces to the caller as the inner exception of a `WasmtimeException`, using the same `CallbackErrorCause` mechanism as host function callbacks, so a host can renew a short deadline per tick and stop the guest on cancellation.

Yielding (`WASMTIME_UPDATE_DEADLINE_YIELD`) is not exposed, as it requires async support this binding does not enable.

Covered by tests for deadline renewal, callback replacement and null arguments.

### DIFF
--- a/src/Store.cs
+++ b/src/Store.cs
@@ -263,6 +263,91 @@ namespace Wasmtime
         /// </summary>
         public void SetData(object? data) => this.data = data;
 
+        /// <summary>
+        /// Represents a callback invoked when the epoch deadline of a store is reached.
+        /// </summary>
+        /// <param name="store">The store whose epoch deadline was reached.</param>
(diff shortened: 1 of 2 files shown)


martindevans (approved):
Looks good, thanks for working on this

PR #375, opened 2026-09-17
Update wasmtime to 48.0.2

phillipcaofph:


phillipcaofph:
I'm keen to get some progress in #346, this seems to a natural first step.

The issue and fix here is found by Claude Opus - hopefully it is correct!

martindevans:
Thanks for looking into this, I investigated the v128 situation myself a while ago and didn't manage to fix it.

I just reviewed the diffs for the c-api from v44 up to v48 and it looks like there are a few other changes that need making before we can make that jump. I'll open an issue with the list of what I found.

martindevans:
I've open a few PRs of my own that fix those changes, the impact was fairly small.

